Issue #13, opened 2020-04-27
Add support for JSON output


coderpatros:
In preparation of v1.2 of the spec we need to support JSON output.

PR #19, opened 2021-05-13
Prepare for JSON BOMs

TedDriggs:
Note: This builds on #18 

This PR splits up the BOM construction to make #13 much easier. The new `Component` struct will have a `Serialize` impl added in a future PR that will emit the JSON format of that BOM element.

This PR also removes all compile-time warnings about unused results and dead code.


stevespringett:
Thanks for the PR @TedDriggs. Is this ready to merge?

TedDriggs:
@stevespringett I ran it on a couple of my projects and it produced equivalent output to before, so I believe it's ready.

PR #20, opened 2021-05-13
Add JSON BOM support

TedDriggs:
* Add `--format` CLI option with `xml` and `json` values
* Render BOM as JSON on request


stevespringett:
Fantastic. Thanks for the PR. Looks like this supports CycloneDX v1.3, correct?

Issue #24, opened 2021-06-25
External reference urls should be inside a <url> tag


sduquette-devolutions:
The URLs are currently placed directly inside the reference tag, they should be inside a <url> tag.

Example:

```
<reference type="documentation">https://docs.rs/base64</reference>
```

instead of

```
<reference type="documentation"><url>https://docs.rs/base64</url></reference>
```

DarthHater:
@sduquette-devolutions I'm going to close this given your PR was merged! Feel free to open if I'm in error

PR #25, opened 2021-06-25
Put external reference urls inside a url tag

sduquette-devolutions:
This is a fix for #24.


stevespringett:
Thanks for finding the issue and submitting a PR. Much appreciated.

PR #33, opened 2021-07-08
Place a single BOM in each crate within a workspace

khuey:
This splits the BOM for a workspace into one for each of its constituent crates.

Cargo's `Workspace::members` appears to return every crate in the workspace, not just the designated `members` in the top-level Cargo.toml, so this drops BOMs in a bunch of library only crates in my workspace at the moment, but it's still a big improvement over the current situation.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -75,25 +75,31 @@ fn real_main(args: Args) -> anyhow::Result<()> {
     let generator = SbomGenerator { all: args.all };
 
     log::trace!("SBOM generation started");
-    let bom = generator.create_sbom(manifest)?;
+    let boms = generator.create_sboms(manifest)?;
     log::trace!("SBOM generation finished");
 
     match args.format {
         Format::Json => {
-            log::info!("Outputting bom.json");
-            serde_json::to_writer_pretty(File::create("bom.json")?, &bom)
-                .map_err(anyhow::Error::from)?;
+            for (manifest_path, bom) in boms {
+                let path = manifest_path.with_file_name("bom.json");
+                log::info!("Outputting {}", path.display());
+                serde_json::to_writer_pretty(File::create(path)?, &bom)
+                    .map_err(anyhow::Error::from)?;
+            }
         }
         Format::Xml => {
-            log::info!("Outputting bom.xml");
-            let file = File::create("bom.xml")?;
-            let file = LineWriter::new(file);
-            let mut xml = XmlWriter::new(file);
+            for (manifest_path, bom) in boms {
+                let path = manifest_path.with_file_name("bom.xml");
+                log::info!("Outputting {}", path.display());
+                let file = File::create(path)?;
+                let file = LineWriter::new(file);
+                let mut xml = XmlWriter::new(file);
 
-            bom.to_xml(&mut xml)?;
-            xml.close()?;
-            xml.flush()?;
-            let _actual = xml.into_inner();
+                bom.to_xml(&mut xml)?;
+                xml.close()?;
+                xml.flush()?;
+                let _actual = xml.into_inner();
+            }
         }
     }
 
--- a/src/generator.rs
+++ b/src/generator.rs
@@ -24,12 +24,15 @@ use cargo::core::Resolve;
 use cargo::core::Workspace;
 use cargo::ops;
 use cargo::Config;
-use std::collections::BTreeSet;
-use std::path::PathBuf;
+use std::collections::{BTreeSet, HashMap};
+use std::path::{Path, PathBuf};
 use thiserror::Error;
 
 pub trait Generator {
-    fn create_sbom<'a>(&self, manifest_path: PathBuf) -> Result<Bom, GeneratorError>;
+    fn create_sboms<'a>(
+        &self,
+        manifest_path: PathBuf,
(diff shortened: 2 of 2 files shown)


amy-keibler:
Hello! Thank you for opening a PR for this. I've reviewed the code and tried it out locally. A few bits of high-level feedback:

1. This no longer merges cleanly with the latest of the main branch. Would you mind resolving the merge conflicts? If not, would you mind if I attempted to do so? (note: much of the affected logic now lives in `src/generator.rs`)
2. Now that we're generating a different BOM for each workspace member, I wonder if the
   ```rust
   // Filter out our own workspace crates from dependency list
   for member in members {
       dependencies.remove(member);
   }
   ```
   snippets should be removed, given that I'd think we want to record inter-project dependencies.
3. I know the code is a bit light on tests at the moment, but if possible it would be great to get some tests around the expected behavior so we don't accidentally break your use-case in the future. A few tests I'd like to see:
   - A test for your expected behavior around a multi-crate workspace
   - A test for the common case of a single crate `Cargo.toml`
4. I would like to allow the user to manually specify the output file path (or even `stdout` maybe) in the future. Do you have any opinion on what the desired behavior would be for a multi-crate workspace in that context? Asking because this change moves us further in the hard-coded direction.

Thank you!

khuey:
>     1. This no longer merges cleanly with the latest of the main branch. Would you mind resolving the merge conflicts? If not, would you mind if I attempted to do so? (note: much of the affected logic now lives in `src/generator.rs`)

Yeah I'll do this when I have some time.

>     2. Now that we're generating a different BOM for each workspace member, I wonder if the
>        ```rust
>        // Filter out our own workspace crates from dependency list
>        for member in members {
>            dependencies.remove(member);
>        }
>        ```
>        snippets should be removed, given that I'd think we want to record inter-project dependencies.

So for my use I actually *don't* want to record the internal dependencies. In my case the entire workspace is a commercial product which is a single unit from the perspective of our customers. None of the internal pieces can be obtained independently, so we don't see any need to expose that information.

It may make sense to make this configurable.
 
>     3. I know the code is a bit light on tests at the moment, but if possible it would be great to get some tests around the expected behavior so we don't accidentally break your use-case in the future. A few tests I'd like to see:
>        
>        * A test for your expected behavior around a multi-crate workspace
>        * A test for the common case of a single crate `Cargo.toml`
> 
>     4. I would like to allow the user to manually specify the output file path (or even `stdout` maybe) in the future. Do you have any opinion on what the desired behavior would be for a multi-crate workspace in that context? Asking because this change moves us further in the hard-coded direction.

I think it would make sense to treat the provided path as a prefix and create <crate-in-the-workspace-name>.xml files for each crate at that prefix. I don't have any idea how you would make it work with stdout.

khuey:
Alright I've rebased this, though I haven't made other changes.

Testing it is kind of involved. We probably need to abstract what `real_main` writes to somehow and have a "fake" output that the test can check.

src/main.rs
```diff
@@ -75,25 +75,31 @@ fn real_main(args: Args) -> anyhow::Result<()> {
...
+    let boms = generator.create_sboms(manifest)?;
     log::trace!("SBOM generation finished");
 
     match args.format {
         Format::Json => {
             log::info!("Outputting bom.json");
-            serde_json::to_writer_pretty(File::create("bom.json")?, &bom)
-                .map_err(anyhow::Error::from)?;
+            for (manifest_path, bom) in boms {
+                let path = manifest_path.with_file_name("bom.json");
```
amy-keibler, line 85: For this and the XML one below, it would be nice to remove the `log::info!` line two lines above and add a line after this one with

``` rust
log::info!("Outputting {}", path.to_string_lossy());
```

amy-keibler (changes requested):
Thanks for rebasing these changes. I tested locally and everything matched my expectations. I have a minor piece of feedback (improving output logging) and the commit needs to be signed off to pass the [DCO check](https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/33/checks?check_run_id=4307053940) to affirm that you have the right to submit these changes and they are licensed under the licenses used by the project ([Apache 2.0](https://github.com/CycloneDX/cyclonedx-rust-cargo/blob/master/LICENSE)).

Once you have the DCO sign-off complete, I'll merge these changes and release a new minor version. Thank you!

src/main.rs
khuey: "Path::display exists for precisely this purpose so I've used that instead."

khuey:
> Once you have the DCO sign-off complete, I'll merge these changes and release a new minor version. Thank you!

Feel free to do that if you want but I also have other PRs on the way :)

amy-keibler (approved):

amy-keibler:
> Feel free to do that if you want but I also have other PRs on the way :)

Sounds good! In case you haven't seen it, I opened #68 and I'm going to be working on implementing it, so some files should be moving around a bit. I'll try to make sure git realizes they are rename operations so it doesn't cause too many conflicts with your branches. Also, if you have any feedback on the proposed design, I'd love to hear it.

khuey:
> Also, if you have any feedback on the proposed design, I'd love to hear it.

I don't see anything discussed there that's obviously a problem :)

PR #43, opened 2021-09-02
Working deploy workflow

DarthHater:
Supercedes #40 , merged everything down into one commit rather than the mess of "LOL how do github actions work"?!

This new workflow can only be triggered manually, and you can specify the release type (major, minor, patch). It bumps the version in the toml file, and also the lock file, and commits this + a new tag of the version to the main branch.

As well, it has the ability to run cargo publish baked in, we just need to turn off dry run and provide a crates token.

I mainly created this so we have a repeatable, transparent release process in the future for the project!

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,41 @@
+name: Deploy to Crates.io
+
+on:
+  workflow_dispatch:
+    inputs:
+      releaseType:
+        description: 'Release type (major, minor, patch)'
+        required: true
+        default: 'patch'
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.4
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+      - name: Cargo bump
+        run: |
+          cargo install cargo-bump
+          cargo bump ${{ github.event.inputs.releaseType }}
+      - name: Retrieve new version
+        run: |
+          echo "::set-output name=CARGO_VERSION::$(grep -e '^version' Cargo.toml | cut -d "=" -f2 | tr -d '/"' | tr -d ' ')"
+        id: version
+      - name: Build one time, for sanity
+        run: cargo build
+      - uses: katyo/publish-crates@v1
+        with:
+          args: --verbose --allow-dirty
+          dry-run: true
+      - name: Configure git and add files
+        run: |
+          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,41 @@
...
+          args: --verbose --allow-dirty
+          dry-run: true
+      - name: Configure git and add files
+        run: |
+          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
+          git commit -am "New development bump to ${{steps.version.outputs.CARGO_VERSION}}"
+          git tag -a ${{steps.version.outputs.CARGO_VERSION}} -m "${{steps.version.outputs.CARGO_VERSION}} release"
+          git push
+          git push origin ${{steps.version.outputs.CARGO_VERSION}}
```
amy-keibler, line 41: I have not tried this locally, so caveat emptor,  but [it looks like there might be a more `git`-y way to express this](https://stackoverflow.com/a/26438076) as `git push --follow-tags` since this tag is annotated.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,41 @@
...
+        run: |
+          cargo install cargo-bump
+          cargo bump ${{ github.event.inputs.releaseType }}
+      - name: Retrieve new version
+        run: |
+          echo "::set-output name=CARGO_VERSION::$(grep -e '^version' Cargo.toml | cut -d "=" -f2 | tr -d '/"' | tr -d ' ')"
+        id: version
+      - name: Build one time, for sanity
+        run: cargo build
+      - uses: katyo/publish-crates@v1
```
amy-keibler, line 30: I did a quick visual audit of their code and didn't see any security concerns with this having access to the crates.io credentials, but we should probably give it a closer look before handing credentials to it.

amy-keibler (approved):

.github/workflows/deploy.yml
DarthHater: I think I avoided that, but in reading I can't figure out why! I'll revise to that.

DarthHater: In 348e136

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          args: --verbose --allow-dirty
+          dry-run: true
```
amy-keibler, line 33:
```suggestion
          registry-token: ${{ secrets.CARGO_API_KEY }}
```

.github/workflows/deploy.yml
amy-keibler: I did a more thorough audit and I feel that we can trust the `v1` version with the token. It appears to pull it in and immediately hand it off to `cargo publish`.

DarthHater:
Yay @amy-keibler !

PR #49, opened 2021-09-30
Address unused CLI arguments

### LINKED ISSUE #48: CLI arguments are not used
The following arguments are supported via the command-line, but are not used:

- `target_dir`
- `verbose`
- `quiet`
- `color`
- `frozen`
- `locked`
- `offline`
- `unstable_flags`
- `config_args`

They are [passed to `cargo::util::Config::configure`](https://github.com/CycloneDX/cyclonedx-rust-cargo/blob/e79a24f81e7a063930434c365764ebf58217733e/src/main.rs#L118), which is only used to get the current working directory in

```rust
let manifest = args
        .manifest_path
        .unwrap_or_else(|| config.cwd().join("Cargo.toml"));
```

I think we probably want to support:

- `verbose`
- `quiet`

and probably drop the rest, as they don't have an effect on SBOM generation (e.g. we're always offline, because we don't download anything from the internet).

We should also consider what other CLI arguments we want to support.

- I think a `--output` would be useful so the user can specify a particular location to store the SBOM rather than defaulting to the current working directory
- Looking at [the Python CycloneDX repository](https://github.com/CycloneDX/cyclonedx-python), I think a `--force` flag would be a good idea to allow users more control on overwriting existing output.
- `--include-dev-dependencies` and `--include-license-text` seem like good options as well, from [the CycloneDX Node repository](https://github.com/CycloneDX/cyclonedx-node-module)

amy-keibler:
Deprecate most of the unused CLI arguments via doc comments and making them private to the `cli` module.

#### Implement the logging CLI arguments of `--quiet` and `--verbose` by pulling in `env_logger` as a dependency

- Cargo does not currently let us express that a dependency is strictly for the binary. We could consider splitting the project into a workspace that contains the CycloneDX specification implementation as a library and a separate project that contains the cargo integration.
- The `cargo` dependency is very noisy, even at `debug`, so I defaulted it to output errors only. Annoyingly, our top level module is `cargo_cyclonedx` and the `env_logger` library uses regex matching, so setting `"cargo"` specifically also impacts our logs.
  - This can be overridden by setting an `RUST_LOG` env variable if you need access to those internal logs (e.g. `RUST_LOG='cargo::=debug' cargo cyclonedx`)
- Currently, there is no log output at the default level (no `--quiet` and no `--verbose` flags), so the observed behavior should be unchanged

#### Clean up

- Remove unused code in `src/bom/metadata.rs` and `src/traits.rs`
- Remove unnecessary `Option` wrappers for `all` and `quiet` arguments. Both should continue to default to `false`
- Update `README.md` to reference the correct executable name

Resolves #48


src/main.rs
```diff
@@ -49,99 +49,43 @@ use cyclonedx_bom::traits::ToXml;
...
-        &args.config_args,
-    )?;
-
-    let manifest = args
-        .manifest_path
-        .unwrap_or_else(|| config.cwd().join("Cargo.toml"));
-    let generator = SbomGenerator {
-        all: Some(args.all),
-    };
+fn real_main(args: Args) -> Result<(), Error> {
```
DarthHater, line 73: lol @ real_main

src/main.rs
```diff
@@ -156,6 +100,47 @@ fn real_main(config: &mut Config, args: Args) -> Result<(), Error> {
...
+
+    builder.parse_default_env(); // allow overriding CLI arguments
+    builder.try_init()?;
+    Ok(())
+}
+
+fn locate_manifest(args: &Args) -> Result<PathBuf, io::Error> {
+    if let Some(manifest_path) = &args.manifest_path {
+        log::info!(
+            "Using manually specified Cargo.toml manifest located: {}",
```
DarthHater, line 130: Minor: maybe `located at:` (reads better, maybe?)

src/main.rs
```diff
@@ -156,6 +100,47 @@ fn real_main(config: &mut Config, args: Args) -> Result<(), Error> {
...
+            manifest_path.to_string_lossy()
+        );
+        Ok(manifest_path.clone())
+    } else {
+        let manifest_path = std::env::current_dir()?.join("Cargo.toml");
+        log::info!(
+            "Using Cargo.toml manifest located: {}",
```
DarthHater, line 137: Minor: maybe located at: (reads better, maybe?)

DarthHater (approved):
LGTM, no major changes other than the deprecations you expressed doing, go for it!

PR #69, opened 2021-11-28
Add repository_url to the purl for non-crates.io packages.

khuey:
For packages that are pulled in by the 
```toml
package = { git = "https://my/git/url" }
```
syntax, include the repository URL in the purl.

### DIFF
--- a/src/component.rs
+++ b/src/component.rs
@@ -227,7 +227,13 @@ impl<'a> From<&'a Package> for ComponentCommon {
     fn from(package: &'a Package) -> Self {
         let name = package.name().to_owned().trim().to_string();
         let version = package.version().to_string();
-        let purl = create_purl(&name, &version);
+        let source_id = package.package_id().source_id();
+        let repository_url = if !source_id.is_default_registry() {
+            Some(source_id.url().as_str())
+        } else {
+            None
+        };
+        let purl = create_purl(&name, &version, repository_url);
 
         Self {
             name,
@@ -247,7 +253,7 @@ impl<'a> From<&'a cargo_metadata::Package> for ComponentCommon {
     fn from(package: &'a cargo_metadata::Package) -> Self {
         let name = package.name.trim().to_string();
         let version = package.version.to_string();
-        let purl = create_purl(&name, &version);
+        let purl = create_purl(&name, &version, None);
 
         Self {
             name,
@@ -258,15 +264,25 @@ impl<'a> From<&'a cargo_metadata::Package> for ComponentCommon {
     }
 }
 
-fn create_purl(name: &str, version: &str) -> Option<String> {
+fn create_purl(name: &str, version: &str, repository_url: Option<&str>) -> Option<String> {
     // PackageUrl::new performs validation on the "type" value passed in. Since we hard-code this value to "cargo", this should always pass.
     // Setting this to gracefully log and handle the potential future case where the "name" field is validated as well.
     // Reference: https://github.com/althonos/packageurl.rs/blob/0.3.0/src/validation.rs#L2
     let purl = PackageUrl::new("cargo", name);
     match purl {
         Ok(mut purl) => {
-            let output = purl.with_version(version.trim()).to_string();
-            Some(output)
+            purl.with_version(version.trim());
+            if let Some(repository_url) = repository_url {
+                if let Err(e) = purl.add_qualifier("repository_url", repository_url) {
+                    log::debug!(
+                        "Error process purl: failed to add qualifier {}: {}",
+                        repository_url,
+                        e
+                    );
+                    return None;
+                }
+            }
+            Some(purl.to_string())
         }
         Err(e) => {
             log::debug!("Error processing purl: {}", e);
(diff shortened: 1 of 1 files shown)


amy-keibler:
I think this is a good addition, but we might need to think through the potential cases slightly. Reading the [documentation for dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations), we can specify multiple ways of acquiring the package (e.g. via a path, via a git repository, and via a crate repository (`crates.io` or other)). I'm not sure offhand what the right approach is here.

For a library, the consumer of the SBOM will likely be resolving the dependency via a crate repository, but for a binary built locally it will be the path or git repository. Maybe the right answer is to include all of the information in the purl and let the user interpret it in the context of the application

khuey:
The purl spec states[0]

> Note: Do not abuse qualifiers: it can be tempting to use many qualifier keys but their usage should be limited to the bare minimum for proper package identification to ensure that a purl stays compact and readable in most cases.
>
>Additional, separate external attributes stored outside of a purl are the preferred mechanism to convey extra long and optional information such as a download URL, vcs URL or checksums in an API, database or web form.
> [...]
> repository_url is an extra URL for an alternative, non-default package repository or registry. When a package does not come from the default public package repository for its type a purl may be qualified with this extra URL. The default repository or registry of a type is documented in the "Known purl types" section.

And for the definition of pkg:cargo[1]

> cargo for Rust:
>
>    The default repository is https://crates.io/

So I think this change, where repository_url is added for crates that come from an alternative registry or from a git url, is correct.

I'll be honest I have no idea what to do about crates that are included from the local filesystem by path, since there's no obvious way to give them a globally unique identifier for the purl, but I don't do that in my product so it isn't a problem for me.

[0] https://github.com/package-url/purl-spec/blob/07d2bdea6d9610b52772866c7ed7859e5566f318/PURL-SPECIFICATION.rst
[1] https://github.com/package-url/purl-spec/blob/07d2bdea6d9610b52772866c7ed7859e5566f318/PURL-TYPES.rst

amy-keibler:
Reading through the documentation and examples, I think the `repository_url` is meant to indicate the root repository source (e.g. `github.com`) based on the uses in [the purl spec's `test-suite-data.json`](https://github.com/package-url/purl-spec/blob/07d2bdea6d9610b52772866c7ed7859e5566f318/test-suite-data.json) and the uses in [other projects in the CycloneDX org](https://github.com/search?q=org%3ACycloneDX+repository_url&type=code).

I think that `vcs_url` (also from [the documentation](https://github.com/package-url/purl-spec/blob/07d2bdea6d9610b52772866c7ed7859e5566f318/PURL-SPECIFICATION.rst#known-qualifiers-keyvalue-pairs)) might be the right solution for `git` dependencies. The format is [defined to be compatible with the SPDX specification](https://github.com/spdx/spdx-spec/blob/cfa1b9d08903/chapters/3-package-information.md#37-package-download-location-). This also appears to be the format used by the [`SourceId`](https://docs.rs/cargo/0.57.0/cargo/core/source/struct.SourceId.html) that we are using in the code change, which leads me to think we are probably on the correct path (We should probably verify this behavior for a variety of sources when testing. I suspect this might also address my concern about there being multiple ways to specify a package dependency in a `Cargo.toml` file).

If you agree with the suggested change to `vcs_url`, I'll pull the code down once it's ready and create an example project that includes all the ways a dependency can be included for testing.

khuey:
Hmm, it's hard to say. `SourceId` in cargo has a `SourceKind` internally that has one of 5 values:

```rust
enum SourceKind {
    /// A git repository.
    Git(GitReference),
    /// A local path.
    Path,
    /// A remote registry.
    Registry,
    /// A local filesystem-based registry.
    LocalRegistry,
    /// A directory-based registry.
    Directory,
}
```

`is_default_registry` returns true if and only if the `SourceId` is `SourceKind::Registry` and the URL is the index for crates.io. It returns false for both other registries (i.e. a crates.io competitor, or an internal corporate registry) and for non-registry things (e.g. `SourceKind::Git`).

For alternate registries I think `repository_url` is correct. In the examples you gave there's `pkg:docker/customer/dockerimage@sha256:244fd47e07d1004f0aed9c?repository_url=gcr.io`. gcr.io as an alternative to the default hub.docker.com is analagous to a crates.io alternative too.

For git repositories I'm less sure but I still think using `repository_url` makes some amount of sense. The crate does not necessarily exist on crates.io, while leaving the `repository_url` unset implies that it does.

There are no `vcs_url` examples in test-suite-data.json so I don't know when to use that.

amy-keibler:
Looking at [Google's documentation](https://cloud.google.com/container-registry/docs/pushing-and-pulling#tag), the form to actually reference images in `gcr.io` is `host/project/image` (with the example they give as `gcr.io/my-project/busybox` in an above section). To me, the equivalent of `gcr.io` in their example is not `https://github.com/CycloneDX/cyclonedx-rust-cargo`, but `github.com`.

They provide an example of `vcs_url` in their [`PURL-TYPES.rst`](https://github.com/package-url/purl-spec/blob/07d2bdea6d9610b52772866c7ed7859e5566f318/PURL-TYPES.rst) file.

```
pkg:generic/bitwarderl?vcs_url=git%2Bhttps://git.fsfe.org/dxtr/bitwarderl%40cc55108da32
```
which decodes to a value of `git+https://git.fsfe.org/dxtr/bitwarderl@cc55108da32`, which appears to be a git repository + commit hash for the release.

I haven't played around with non-crates.io registries, but my best guess so far based on the documentation is that we want the behavior to populate `vcs_url` in the case of `SourceKind::Git` (in the `<vcs_tool>+<transport>://<host_name>[/<path_to_repository>][@<revision_tag_or_branch>][#<sub_path>]` form if we don't get that for free from cargo), `repository_url` in the case of non-crates.io `SourceKind::Registry`, and do nothing yet for the other cases.

Shnatsel:
This PR is superseded by #523, which is now merged. Closing.

Issue #93, opened 2022-01-19
Replace StructOpt with clap


amy-keibler:
We use [`StructOpt`](https://crates.io/crates/structopt) for our CLI parser library. Its functionality has been added to one of its dependencies, [`clap`](https://crates.io/crates/clap) in its `3.0.0` release. `StructOpt` is now in [maintenance mote](https://github.com/TeXitoi/structopt#maintenance), so we should switch over to `clap`. This should mostly translate directly, but there might be a few changes required

### Acceptance Criteria

- [ ] The `structopt` dependency is replaced with `clap = "3.0.10"`
- [ ] Existing CLI functionality is preserved or improved

allisonsierra:
I will work on this one.

PR #94, opened 2022-01-20
Replace StructOpt with clap

allisonsierra:
Implements #93

Paired with @amy-keibler


amy-keibler (approved):
Paired on testing the changes locally and no functionality is lost or modified with these changes ✅

PR #103, opened 2022-02-03
Add ability to configure via Cargo.toml metadata

### LINKED ISSUE #74: Use Cargo.toml metadata for command inputs
The [Cargo manifest format](https://doc.rust-lang.org/cargo/reference/manifest.html#the-metadata-table) supports a `[package.metadata]` table to be used by external tools. We should allow a

``` toml
[package.metadata.cyclonedx]
all = true
format = "json"
```
style section. This should allow the user to specify how they want their SBOM files to be generated in a consistent manner without requiring them to specify CLI arguments every time.

`manifest-path` doesn't make sense to support here, because we are reading the values from the manifest. Likewise, `quite` and `verbose` control CLI behavior and make sense to be specified by the specific CLI invocation.

### Acceptance Criteria

- [ ] A user can configure their root-level `Cargo.toml` file with a relevant `package.metadata.cyclonedx` section and specify configurations for `all` and `format`
- [ ] CLI arguments override a configuration (e.g. a `Cargo.toml` with `format = "json"`, but a CLI invocation of `cargo cyclonedx --format xml` should produce XML output
- [ ] The configuration is documented in the `README.md`

allisonsierra:
Add GeneratedSbom struct and moved file output responsbility to it

Resolves #74

Paired with @amy-keibler

Signed-off-by: Allie Sierra <85902135+allisonsierra@users.noreply.github.com>


amy-keibler (approved):
Tested locally ✅

PR #243, opened 2022-08-31
Integrate BOM library with CLI

allisonsierra:
- Adds needed constructors and trait implementations to BOM library
- Changes Purl to hold a String instead of a Uri


cargo-cyclonedx/Cargo.toml
```diff
@@ -24,17 +24,14 @@ lto = true
 [dependencies]
 anyhow = "1.0.62"
 cargo = "0.64.0"
-cargo_metadata = "0.15.0"
 clap = { version = "3.2.17", features = ["derive"] }
+cyclonedx-bom = { path = "../cyclonedx-bom" }
```
amy-keibler, line 28: We'll need a follow-up PR to set the version explicitly before we publish to crates.io. We'll also need to update the deployment process to automatically bump the version here when we publish a new release of the library (some investigation required to see how other projects handle this)

cargo-cyclonedx/src/author.rs
```diff
@@ -36,11 +33,36 @@ pub enum AuthorParseError {
     Email(String),
 }
 
-#[derive(Debug, Serialize, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug)]
+pub struct Authors(pub Option<Vec<Author>>);
```
amy-keibler, line 37: We'll need a follow-up issue to track authors:

- Drop the email regex and replace it with a dependency?
- Better handling of invalid authors

cargo-cyclonedx/src/generator.rs
```diff
@@ -83,9 +94,9 @@ impl SbomGenerator {
                     top_level_dependencies(&members, &package_ids)?
                 };
 
-            let mut bom: Bom = dependencies.iter().collect();
+            let bom = create_bom(member, dependencies);
 
-            bom.metadata = get_metadata(member.manifest_path());
+            log::debug!("Bom validation: {:?}", &bom.validate());
```
amy-keibler, line 99: Should we drop this line before merging?

cyclonedx-bom/src/models/metadata.rs
```diff
@@ -38,6 +38,21 @@ pub struct Metadata {
     pub properties: Option<Properties>,
 }
 
+impl Metadata {
+    pub fn new() -> Self {
+        Self {
+            timestamp: DateTime::now().ok(),
```
amy-keibler, line 44: We should probably have an issue to figure out logging and link this to it so we don't silently hide timestamp issues

cyclonedx-bom/src/external_models/uri.rs
```diff
@@ -16,16 +16,32 @@
...
     FailureReason, Validate, ValidationContext, ValidationError, ValidationResult,
 };
 
-#[derive(Debug, PartialEq, Eq)]
-pub struct Purl(pub(crate) Uri);
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Purl(pub(crate) String);
+
+impl Purl {
+    pub fn new(name: &str, version: &str, package_type: &str) -> Result<Purl, UriError> {
```
amy-keibler, line 32: Style preference: I'd probably have `package_type` first, since that's the order it shows up in a purl.

amy-keibler (approved):

cargo-cyclonedx/src/generator.rs
allisonsierra: While we in theory should be constructing a valid BOM, I like being able to prove that with the library. This did expose some problems when initially doing the integration so it's very helpful as a developer.

I think it makes sense to leave this since it's behind a log level switch. We could move it to `log::trace` but I think debug makes more sense because trace is really noisy.

PR #294, opened 2022-10-13
Add docs to organization.rs

allisonsierra:
Add docs to `organization.rs`

### DIFF
--- a/cyclonedx-bom/src/models/organization.rs
+++ b/cyclonedx-bom/src/models/organization.rs
@@ -23,6 +23,9 @@ use crate::{
     },
 };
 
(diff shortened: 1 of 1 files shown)


cyclonedx-bom/src/models/organization.rs
```diff
@@ -73,6 +82,9 @@ impl Validate for OrganizationalContact {
     }
 }
 
+/// Represents an organization with name, url and contact information
```
amy-keibler, line 85:
```suggestion
/// Represents an organization with name, url, and contact information
```

amy-keibler (approved):

PR #295, opened 2022-10-13
Add docs to metadata.rs

allisonsierra:
Add docs to metadata.rs

### DIFF
--- a/cyclonedx-bom/src/models/metadata.rs
+++ b/cyclonedx-bom/src/models/metadata.rs
@@ -28,6 +28,9 @@ use crate::validation::{
     Validate, ValidationContext, ValidationError, ValidationPathComponent, ValidationResult,
 };
 
+/// Represents additional information about a BOM
+///
+/// Defined via the [CycloneDX XML schema](https://cyclonedx.org/docs/1.3/xml/#type_metadata)
 #[derive(Debug, Default, PartialEq, Eq)]
 pub struct Metadata {
     pub timestamp: Option<DateTime>,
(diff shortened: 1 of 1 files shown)


cyclonedx-bom/src/models/metadata.rs
```diff
@@ -41,6 +44,15 @@ pub struct Metadata {
 }
 
 impl Metadata {
+    /// Constructs a new `Metadata` with a timestamp of DateTime::now()
+    /// ```
+    /// use cyclonedx_bom::models::metadata::Metadata;
+    ///
+    /// let metadata = Metadata::new();
```
allisonsierra, line 51: This isn't a very good example since this returns a result, so open to feedback here. In `generator.rs` we are using the `?` operator but can't do that in the doctest as it's currently written.

cyclonedx-bom/src/models/metadata.rs
```diff
@@ -41,6 +44,15 @@ pub struct Metadata {
...
+    /// ```
+    /// # Errors
+    ///
+    /// Returns an error variant if unable to generate a valid timestamp based on the current time
```
allisonsierra, line 55: Also wondering if this could use a better description. Based off the Rust API doc style guide.

cyclonedx-bom/src/models/metadata.rs
```diff
@@ -41,6 +44,15 @@ pub struct Metadata {
...
 impl Metadata {
+    /// Constructs a new `Metadata` with a timestamp based on the current time
+    /// ```
+    /// use cyclonedx_bom::models::metadata::Metadata;
+    ///
+    /// let metadata = Metadata::new();
+    /// ```
+    /// # Errors
+    ///
+    /// Returns an error variant if unable to generate a valid timestamp
```
allisonsierra, line 55: I'm open to suggestions on this description. It's based on the Rust API docs style guide.

cyclonedx-bom/src/models/metadata.rs
allisonsierra: Went with this style: https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#using--in-doc-tests

amy-keibler (approved):

PR #365, opened 2023-01-09
Lookup top level dependencies also from resolved dependency graph

### LINKED ISSUE #364: Using a git patch on a dependecy removes it from the SBOM
Assume the following `Cargo.toml`:

```patch
[package]
name = "cyclonedx-rust-repro-1"
version = "0.1.0"
edition = "2021"

[dependencies]
base64 = "0.21.0"

+[patch.crates-io]
+base64 = { git = "https://github.com/marshallpierce/rust-base64" }
```

When the patch is missing, the bom renders as:

```xml
  <components>
    <component type="library" bom-ref="pkg:cargo/base64@0.21.0">
      <name>base64</name>
      <version>0.21.0</version>
      <description>encodes and decodes base64 as bytes or utf8</description>
      <scope>required</scope>
      <licenses>
        <expression>MIT OR Apache-2.0</expression>
      </licenses>
      <purl>pkg:cargo/base64@0.21.0</purl>
      <externalReferences>
        <reference type="documentation">
          <url>https://docs.rs/base64</url>
        </reference>
        <reference type="vcs">
          <url>https://github.com/marshallpierce/rust-base64</url>
        </reference>
      </externalReferences>
    </component>
  </components>
```

When the patch is enabled, it is missing:

```xml
  <components />
```

ctron:
Instead of using the declared dependency, this change now uses to the resolved dependency graph to look up the top level dependencies.

The difference is, that a declared dependency may be overridden. And so the source no longer matches, which results in the dependency missing from the final SBOM.

Using the resolved dependency graph, overriding dependencies is covered as well.

This also aligns with the "all dependencies" approach, which also uses the resolved dependencies.

Closes #364

### DIFF
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -23,6 +23,7 @@ use crate::format::Format;
 use crate::toml::config_from_toml;
 use crate::toml::ConfigError;
 use cargo::core::dependency::DepKind;
+use cargo::core::Dependency;
 use cargo::core::Package;
 use cargo::core::PackageSet;
 use cargo::core::Resolve;
@@ -91,7 +92,7 @@ impl SbomGenerator {
                 if config.included_dependencies() == IncludedDependencies::AllDependencies {
                     all_dependencies(&members, &package_ids, &resolve)?
                 } else {
-                    top_level_dependencies(&members, &package_ids)?
+                    top_level_dependencies(&members, &package_ids, &resolve)?
                 };
 
             let bom = create_bom(member, dependencies)?;
@@ -374,14 +375,24 @@ pub enum GeneratorError {
 fn top_level_dependencies(
     members: &[Package],
     package_ids: &PackageSet<'_>,
+    resolve: &Resolve,
 ) -> Result<BTreeSet<Package>, GeneratorError> {
     log::trace!("Adding top-level dependencies to SBOM");
     let mut dependencies = BTreeSet::new();
 
     let all_dependencies = members
         .iter()
-        .flat_map(|m| m.dependencies().iter())
-        .filter(|d| d.kind() == DepKind::Normal);
+        .flat_map(|m| {
+            resolve
+                .deps(m.package_id())
+                .filter_map(move |r| match r.0 == m.package_id() {
+                    true => Some(r.1),
+                    false => None,
+                })
+        })
+        .flatten()
+        .filter(|d: &&Dependency| d.kind() == DepKind::Normal);
+
     for dependency in all_dependencies {
         log::trace!("Dependency: {dependency:?}");
         match package_ids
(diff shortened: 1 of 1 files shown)


cargo-cyclonedx/src/generator.rs
```diff
@@ -374,23 +375,44 @@ pub enum GeneratorError {
...
         .iter()
-        .flat_map(|m| m.dependencies().iter())
-        .filter(|d| d.kind() == DepKind::Normal);
+        .flat_map(|m| {
+            resolve
+                .deps(m.package_id())
+                .filter_map(move |r| match r.0 == m.package_id() {
+                    true => Some(r.1),
+                    false => None,
+                })
```
amy-keibler, line 391: Style preference: In my opinion, this is cleaner as the following and likely can be optimized to similar instructions.

```suggestion
                .filter(move |r| r.0 == m.package_id())
                .map(|(_, dependency)| dependency)
```

Not required, but thought it was worth suggesting

amy-keibler (approved):
Thanks for identifying and fixing this issue!

For CI, if you rebase off the latest `main`, I've pushed changes that address the new clippy lints that were failing. Once CI is passing, I'll merge this in.

cargo-cyclonedx/src/generator.rs
ctron: Works for me, I don't have any strong opinion there.

ctron:
Accepted the code changes and re-signed-off the commit

lfrancke:
I'm afraid this broke the top-level dependency stuff entirely.

0.3.8 doesn't produce a component list at all anymore.

I added debugging like this:
```
                .filter(move |r| {
                    println!("{} == {}", r.0, m.package_id());
                    r.0 == m.package_id()
                })
```

and if I try to generate a BOM for the `cyclonedx-bom` project itself this is what I get:

```
cyclonedx-bom v0.4.1 (/home/lars/dev/external/cyclonedx-rust-cargo/cyclonedx-bom) == cargo-cyclonedx v0.3.8 (/home/lars/dev/external/cyclonedx-rust-cargo/cargo-cyclonedx)
```

I tried various other projects and all of them return an empty component list.
The same works on 0.3.7.

lfrancke:
@ctron - if you have a few moments to take a look at this, that'd be great

ctron:
I am sorry, but I can't reproduce this. I did update to the most recent `main` branch, and ran:

```
cargo run -- cyclonedx -f json  -v
```

Which gives me two SBOMs:

```
cargo-cyclonedx/bom.json
cyclonedx-bom/bom.json
```

Both look reasonable. Also adding `--all` seems to work just fine.

I am using:

```
cargo 1.71.0 (cfd3bbd8f 2023-06-08)
rustc 1.71.0 (8ede3aae2 2023-07-12)
```

lfrancke:
It was fixed in main on Friday.
Could you try the commit 9cb6d175897c8ecfe3bd0466b73468450cd5b815?

PR #376, opened 2023-02-09
fix visibility of `HashValue`

jondot:
This enables to actually provide hashes

### DIFF
--- a/cyclonedx-bom/src/models/hash.rs
+++ b/cyclonedx-bom/src/models/hash.rs
@@ -158,7 +158,7 @@ impl Validate for HashAlgorithm {
 
 /// Defined via the [CycloneDX XML schema](https://cyclonedx.org/docs/1.3/xml/#type_hashValue)
 #[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


lfrancke:
Could you add the DCO to your commit?

See here for details: https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/376/checks?check_run_id=11218823763

Shnatsel:
The ability to write the hash is definitely required, but I don't think this is the way to do. I would prefer to expose a method that accepts a string and validates it, and only creates the hash struct if the string is actually valid for the supplied algorithm.

Shnatsel:
Nevermind, I take that back. 

Introducing a type-level distinction between validated and non-validated data would be great, but would require a rather significant refactoring of the whole crate for consistency. It does not make sense to introduce it in some places but not others.

lfrancke:
Validating hashes seems like a nice to have but definitely also not trivial to do and not sure if we even want to get into that...

lfrancke:
See #519 for a new version of this.
I'm going to close this in favor of the new one.
Thank you for bringing it up!

PR #381, opened 2023-03-23
Expose more Details of Purl and Cpe for writing

koa:


lfrancke:
Hi,
would you mind adding your sign off/DCO as per these instructions?

https://github.com/CycloneDX/cyclonedx-rust-cargo/runs/12224035251

Thank you

cyclonedx-bom/src/external_models/uri.rs
```diff
@@ -60,6 +60,12 @@ impl Validate for Purl {
     }
 }
 
+impl<'a> From<PackageUrl<'a>> for Purl {
```
Shnatsel, line 63: The PackageUrl crate currently has version 0.3.0, not 1.0 or something like that. It is likely that they will bump the version frequently. If we provide a `From<PackageUrl>`  implementation, that will expose the `PackageUrl` type in the public API, and require an incompatible SemVer change to this crate every time the SemVer of `PackageUrl` is bumped.

Could you provide a `FromStr` implementation instead?

Shnatsel (changes requested):
Needs one change, other than that looks good. Thanks!

Shnatsel (approved):

Shnatsel:
Thanks! Just needs a signoff with `git rebase HEAD~5 --signoff`  and then `git push --force`, and I can merge!

PR #481, opened 2023-10-06
Init release pipeline

keponk:
Update cyclonedx-bom CD to build release binary, zip and release in github.


lfrancke:
FYI @Shnatsel is planning improvements here as well

Issue #474, opened 2023-09-29
Getting different behavior between repo and cargo install


keponk:
Hi, I have a small project I was trying to run this on. A dummy one with only a handful of dependencies.

I installed the tool with cargo and as per the generated bom it is the `0.3.8` version. But this is presenting the exact problem stated here: https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/443  I don't see the dependencies in the resulting bom file.

I noticed this is supposed to be fixed, and so I tried cloning the repo, compiling and ran the resulting release build, this worked and didn't present the problem shown above. I can see the version is also shown as `0.3.8`. So I'm wondering if there's some mismatch between the version installed by `cargo install` and what is currently on the main branch.

lfrancke:
Hi, 
I'm sorry I'm not entirely sure I understand.

- The released version is 0.3.8 and that doesn't work
- The current main branch works for you?

lfrancke:
Understood! Thank you for the clarification.

When you do `cargo install` you'll get the latest published version which is 0.3.8: https://crates.io/crates/cargo-cyclonedx

Unfortunately, no release has been made with the fixes yet. I'll see what we can do about it.

keponk:
thanks! I also noticed there's no automated cicd here and couldn't find documentation around it. I'm wondering if you have a reason to not have it automated, in other words, that the releases are accompanied by their compiled release target in the 2 or 3 major architectures.
The underlying reason for me is that I'm trying to also use this in my CICD, and getting the cargo version, besides being currently broken, also needs compilation during install, if I could just pull the "official" binary release that would be awesome.
I would be happy to try putting something together but  am wondering if you have your reasons to intentionally not do that.

lfrancke:
This happened before my time but the quick answer is probably because no one has had the time so far :)

You are looking for releases in GitHub, right?

keponk:
Github or you preferred location, I assumed github is the expected location.
Bottom line, the ultimate goal is to generate SBOMs right? this are usually generated during CI process.
So the ability to quickly grab that binary to do it's job during CI would be ideal. Currently the alternatives seem to be
1. cargo install : besides the issue of not having the latest version in it, it takes time to compile
2. Some more convoluted linking to this repo / checkout and also compiling.

The ideal scenario could be to have away to quickly grab/download the release binary and then project wanting to use it during CI can simply run that binary. No compilation required.

lfrancke:
[...]

PR #488, opened 2023-10-19
Add prod build and release steps

keponk:
https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/474
This one might require a closer review. A quick summary:

This PR adds 2 new jobs in the same deploy_cargo action. The existing job which tags and publishes has zero changes.

The first new one is `build-prod` running _after_ `deploy` has completed. This simply runs the build process matrix with the `--release --lock` tag, creating a smaller and optimised binary. Along with this it also renames the binary when necessary (the hypothetical case of a windows `.exe` existing), then it zip-packages the file (the size difference it quite noticeable), and creates the sha256 for that packaged file. These files are then uploaded as artifacts.

The second job, `release` also depends on the previous one, it collects all the artifacs to release them together and names the release using the _tag_ name.

Other remarks:
I tested this on my local repo and was able to get the _latest_ release from the release public url, so at least for my niche use case it works :) . The windows build always fails with some error which was a bit above my head to debug, but I will assume this is not a local issue. If that's the case then this tool currently doesn't support windows, and maybe is worth adding that comment on the README.
Finally, in the process of learning about "best practices" I noticed these 2 projects : [cargo-binstall](https://crates.io/crates/cargo-binstall) and [cargo-quickinstall](https://github.com/cargo-bins/cargo-quickinstall) both rely on a combination of pre-compiled binaries by them (for the more popular packages) and some crawling on the crate repo to find latest releases. So by having this release, I believe you also become automatically supported by this projects 👍

### DIFF
--- a/.github/workflows/deploy_cargo_cyclonedx.yml
+++ b/.github/workflows/deploy_cargo_cyclonedx.yml
@@ -11,6 +11,8 @@ on:
 jobs:
   deploy:
     runs-on: ubuntu-latest
+    outputs:
+      TAG_NAME: ${{ steps.version.outputs.TAG_NAME }}
     steps:
       - name: JQ
         run: |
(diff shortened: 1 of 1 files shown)


Shnatsel:
Thank you for your contribution!

I am hoping that https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/496 will fix the Windows build by dropping the C system dependencies. I'll let you know once it's merged so you could rebase and try Windows again.

Shnatsel:
https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/496 is now merged, so C dependencies are gone. Could you try again and see if this fixed the Windows release workflow?

keponk:
@Shnatsel indeed, the windows build is now succeeding.

I've updated this PR and tested on my local setup. The PR description is still valid except that now it covers the 3 OSs: win, linux, macos

Shnatsel:
Looks good. Could you also document the expected tag format? Or will it trigger for literally any tag?

keponk:
@Shnatsel this is "tied" to the existing deploy file. So in this scenario, as longs as you've decided to _tag_ anything, it is also getting released. This can be changed of course. But the behavior is straightforward:
tag&publish > build prod > release

in other words, if you've already taken the decision to _publish_ a new version (this is a manually triggered action) then it is also creating a release.

Shnatsel:
Ah, excellent. Good to go then!

lfrancke:
Thank you very much for your contribution!

Shnatsel:
The first release build with this workflow is now up: https://github.com/CycloneDX/cyclonedx-rust-cargo/releases/tag/cargo-cyclonedx-0.4.0

Thanks!

Issue #495, opened 2023-10-26
SBOM configuration via `Cargo.toml` appears harmful


Shnatsel:
Right now `cargo-cyclonedx` will read some of its configuration from the `Cargo.toml` of the project it is run on. That is, the developers rather the user who runs the tool determine what kind of output it will produce.

For example, if the user expects XML files (as per the default) but the project contains a directive to output JSON, the configuration in `Cargo.toml` will override the default, and an unusable SBOM will be generated.

It will be even worse if we default to reporting all dependencies in the future, but a `Cargo.toml` file specifies reporting only direct dependencies; in this case the SBOM will silently be incomplete, creating security and compliance issues.

I believe the possibility of the configuration to be specified by anyone other than the person actually using the tool is a fundamentally flawed approach, and all such means of configuration should be removed.

@lfrancke thoughts?

lfrancke:
I tend to agree with you. Good catch! I didn't even think of that.

So, the suggestion is to remove the option to configure via `Cargo.toml` entirely? I think I'm good with that. I'll ask in Slack if anyone has any opinions on this but if we don't hear anything in a few days I'd say we can remove it.

Shnatsel:
Removed in #520

Issue #489, opened 2023-10-23
Download crates in parallel


Shnatsel:
`cargo cyclonedx` downloads crates from crates.io when they are not cached locally.

The download is sequential, that is crates are downloaded one by one, and the next download does not start until the previous one completes. This takes in the ballpark of a second per crate due to network latency and several roundtrips involved, which quickly adds up.

Downloading crates in parallel like `cargo` itself does would be much faster.

There are three possible ways to go about it:

1. Just delegate to `cargo fetch`
2. Use something like the [`cargo-fetcher`](https://crates.io/crates/cargo-fetcher) crate, but that would add a lot of dependencies and internal complexity (e.g. an async runtime)
3. Just hand-roll the download, it is fairly trivial after all.

I believe only option 1 would correctly handle things like SOCKS proxies, and we should use it given the opportunity.

Shnatsel:
Fixed by #496

PR #496, opened 2023-10-26
Convert to `cargo metadata` as a backend

Shnatsel:
This PR switches to the API-stable `cargo metadata` as opposed to the previous direct dependency on unstable internal APIs of Cargo.

It is mostly bug-compatible, with the actual fixes to come in subsequent PRs.

### Benefits

 - Far smaller binary size
 - Shorter compile times
 - Reliance only on stable APIs will require far less maintenance going forward
 - https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/489 (we got it for free by switching to `cargo metadata`)

### Fixes

One of the tests was creating an invalid dependency that was getting ignored by Cargo, but still included in the SBOM. This is now fixed, and I've updated the test to create a proper dependency.

### Behavior changes

The previous code was inconsistent on what kinds of dependencies it includes. It previously included normal only (not build!) in top-level only mode, and all kinds (even dev!) in all deps mode. Now all kinds of dependencies are included always - it's still wrong, but at least it's consistent. Proper filtering by mode and package is implemented in a subsequent PR: #498

### Regressions

Reading configuration from `Cargo.toml` is not fully wired up, although its tests pass. I've skipped it for now and left a comment because this behavior seems to be a bad idea in the first place: https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/495

### DIFF
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -87,17 +87,14 @@ impl SbomGenerator {
             log::trace!("Config from config override: {:?}", config_override);
             log::debug!("Config from merged config: {:?}", config);
 
-            // TODO: restore support for reporting top-level dependencies only
-            // (assuming that mode is compliant with the CycloneDX spec)
+            let (dependencies, resolve) =
+                if config.included_dependencies() == IncludedDependencies::AllDependencies {
+                    all_dependencies(member, &packages, &resolve)
+                } else {
+                    top_level_dependencies(member, &packages, &resolve)
+                };
 
-            // let dependencies =
-            //     if config.included_dependencies() == IncludedDependencies::AllDependencies {
-            //         all_dependencies(&members, &package_ids, &resolve)?
-            //     } else {
-            //         top_level_dependencies(member, &package_ids, &resolve)?
-            //     };
-
-            let bom = create_bom(member, &packages)?;
+            let bom = create_bom(member, &dependencies)?;
 
             log::debug!("Bom validation: {:?}", &bom.validate());
 
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -88,7 +88,7 @@ impl SbomGenerator {
             log::trace!("Config from config override: {:?}", config_override);
             log::debug!("Config from merged config: {:?}", config);
 
-            let (dependencies, resolve) =
+            let (dependencies, _resolve) =
                 if config.included_dependencies() == IncludedDependencies::AllDependencies {
                     all_dependencies(member, &packages, &resolve)
                 } else {
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -29,6 +29,10 @@ use cargo::core::Resolve;
 use cargo::core::Workspace;
 use cargo::ops;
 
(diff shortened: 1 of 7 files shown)


lfrancke:
Thank you for the PR!

I did take a quick look and have one initial question: We currently run `cargo metadata` once for everything. As far as I know it can be run against each package individually.
Wouldn't that solve the issues we currently have?
Yes, it'd mean running cargo metadata multiple times so it'd be a bit slower but that's a tradeoff I'd be happy to take.

Disclaimer: I have not looked at your follow-up PRs yet.

cargo-cyclonedx/src/generator.rs
```diff
@@ -372,65 +388,44 @@ pub enum GeneratorError {
 }
 
 fn top_level_dependencies(
-    member: &Package,
-    package_ids: &PackageSet<'_>,
-    resolve: &Resolve,
-) -> Result<BTreeSet<Package>, GeneratorError> {
+    member: &PackageId,
```
lfrancke, line 391: This seems to be unused now. Rust is complaining.

cargo-cyclonedx/src/generator.rs
```diff
@@ -372,65 +388,44 @@ pub enum GeneratorError {
...
+    pkg_result.insert(member.to_owned(), packages[member].to_owned());
+    for id in direct_dep_ids {
+        pkg_result.insert(id.to_owned(), packages[id].to_owned());
     }
 
-    Ok(dependencies)
+    let mut resolve_result = ResolveMap::new();
+    resolve_result.insert(member.to_owned(), resolve[member].clone());
+    for id in direct_dep_ids {
+        // Clear all depedencies, pretend there is only one level
```
lfrancke, line 407:
```suggestion
        // Clear all dependencies, pretend there is only one level
```

Shnatsel:
Even if we ran it for every package separately, we would still have to do the graph traversal to filter out the transitive dev-dependencies. That's implemented in #498. So running `cargo metadata` per-package doesn't really get us anything.

cargo-cyclonedx/src/generator.rs
```diff
@@ -87,21 +88,21 @@ impl SbomGenerator {
             log::trace!("Config from config override: {:?}", config_override);
             log::debug!("Config from merged config: {:?}", config);
 
-            let dependencies =
+            let (dependencies, _resolve) =
```
lfrancke, line 91: Is this unused in preparation for the dependency tree?

lfrancke (changes requested):
Minor stuff
It's really really hard to not comment on some of the obvious wrong things :)

cargo-cyclonedx/src/generator.rs
Shnatsel: This is unused for now, but subsequent PRs make it used again. This is definitely the signature we'll need.

cargo-cyclonedx/src/generator.rs
Shnatsel: Exactly.

cargo-cyclonedx/src/generator.rs
lfrancke: I assumed as much but since you used the underscore for the `_resolve` but not here, I wasn't sure.
We can keep it as it is

lfrancke (approved):

Issue #444, opened 2023-07-30
"All" dependencies include all dependencies across all packages


lfrancke:
As the title says.
If I have a project with multiple packages (e.g. this one, `cyclonedx-rust-cargo`) then the components for all packages will include _all_ components from all packages combined.

PR #498, opened 2023-10-26
Accurate deps

Shnatsel:
**Depends on #496, please review that first!**

- Makes reporting more accurate by excluding dev-dependencies, which cannot affect the production library/binary/etc by design.
    - A flag for including dev-dependencies of the toplevel package only can be added later, if desired, but transitive dev-dependencies definitely should always be omitted.
- Only includes the dependencies of the specified package, fixing #444

### DIFF
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -87,17 +87,14 @@ impl SbomGenerator {
             log::trace!("Config from config override: {:?}", config_override);
             log::debug!("Config from merged config: {:?}", config);
 
-            // TODO: restore support for reporting top-level dependencies only
-            // (assuming that mode is compliant with the CycloneDX spec)
+            let (dependencies, resolve) =
+                if config.included_dependencies() == IncludedDependencies::AllDependencies {
+                    all_dependencies(member, &packages, &resolve)
+                } else {
+                    top_level_dependencies(member, &packages, &resolve)
+                };
 
-            // let dependencies =
-            //     if config.included_dependencies() == IncludedDependencies::AllDependencies {
-            //         all_dependencies(&members, &package_ids, &resolve)?
-            //     } else {
-            //         top_level_dependencies(member, &package_ids, &resolve)?
-            //     };
-
-            let bom = create_bom(member, &packages)?;
+            let bom = create_bom(member, &dependencies)?;
 
             log::debug!("Bom validation: {:?}", &bom.validate());
 
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -427,7 +427,7 @@ fn all_dependencies(
 ) -> (PackageMap, ResolveMap) {
     log::trace!("Adding all dependencies to SBOM");
 
-    // Note: using Vec (without deduplication) can theoreticall cause quadratic memory usage,
+    // Note: using Vec (without deduplication) can theoretically cause quadratic memory usage,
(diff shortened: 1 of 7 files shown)


cargo-cyclonedx/src/generator.rs
```diff
@@ -388,44 +390,93 @@ pub enum GeneratorError {
...
+    for id in &root_node.dependencies {
+        // Clear all depedencies, pretend there is only one level
         let mut node = resolve[id].clone();
         node.deps = Vec::new();
         node.dependencies = Vec::new();
-        resolve_result.insert(id.to_owned(), node);
+        resolve_result.insert((*id).to_owned(), node);
     }
+    // Insert the root node at the end now that we're done iterating over it
+    resolve_result.insert(root.to_owned(), root_node);
```
lfrancke, line 418: Does this mean we'll add ourselves to our own dependencies or is this (again) only used for the dependency tree later?

cargo-cyclonedx/src/generator.rs
```diff
@@ -388,44 +390,93 @@ pub enum GeneratorError {
...
+    root: &PackageId,
     packages: &PackageMap,
     resolve: &ResolveMap,
 ) -> (PackageMap, ResolveMap) {
     log::trace!("Adding all dependencies to SBOM");
 
-    // FIXME: run BFS to filter out irrelevant dependencies,
-    // such as dev dependencies that do not affect the final binary
-    // or dependencies of other packages in the workspace
+    // Note: using Vec (without deduplication) can theoreticall cause quadratic memory usage,
```
lfrancke:
```suggestion
    // Note: using Vec (without deduplication) can theoretically cause quadratic memory usage,
```

cargo-cyclonedx/src/generator.rs
```diff
@@ -388,44 +390,93 @@ pub enum GeneratorError {
...
+    // but since `Node` does not implement `Ord` or `Hash` it's hard to deduplicate them.
+    // These are all pointers and there's not a lot of them, it's highly unlikely to be an issue in practice.
+    // We can work around this by using a map instead of a set if need be.
+    let mut current_queue: Vec<&Node> = vec![&resolve[root]];
+    let mut next_queue: Vec<&Node> = Vec::new();
+
+    let mut out_resolve = ResolveMap::new();
+
+    // Run breadth-first search (BFS) over the dependency graph
+    // to determine which nodes are actually dependened on by our package
```
lfrancke, line 440:
```suggestion
    // to determine which nodes are actually depended on by our package
```

cargo-cyclonedx/src/generator.rs
```diff
@@ -388,44 +390,93 @@ pub enum GeneratorError {
...
+    // (not other packages) and to remove dev-dependencies
+    while current_queue.len() != 0 {
```
lfrancke, line 442:
```suggestion
    while !current_queue.is_empty() {
```

lfrancke (changes requested):
Only minor stuff again. I tested it against itself and it worked just fine, tests ran as well.

I'll run it against one of our projects now to double check but don't expect any issues.

cargo-cyclonedx/src/generator.rs
Shnatsel: This is only used for the dependency tree. We do not add the toplevel package to `components`, preserving the original behavior of the tool.

lfrancke (approved):

lfrancke:
Tested it against two of our own projects and it seems to fix #444
Output looks good from my spot checks.

PR #499, opened 2023-10-26
Slience lax parsing "errors"

Shnatsel:
They are not, in fact, errors. Just a conversion from the Rust format to the more universal one.

Supersedes #363


lfrancke:
This looks good to me but I want to give others a few days to voice their opinions as well.
Especially @amy-keibler @ctron and @sebsnyk as they were part of the previous #363  effort and this does something different.
So, I'll wait for at least one week before merging this. Feedback welcome.

I also think part of #363 is not implemented here, namely the option to enable lax parsing or am I mistaken?

lfrancke:
Makes sense. I'm also fine with adding #363 instead. Would you be willing to update that and we can work on getting it in?

Shnatsel:
Closing in favor of https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/363

Issue #409, opened 2023-05-30
Make dependency content public.


alilleybrinker:
As far as I can tell, there's no valid way to extract the dependency information in a `Bom`. The `dependencies` field is public, but it's a newtype struct with a `pub(crate)` field, meaning it's private for other crates. There are no methods to access the internal `Vec` of dependencies, so there's no way to access this field.

I suppose there's a workaround of using the JSON output to get a `serde_json::Value` and then using the `pointer` method to access the field, but this feels excessive for accessing data that's present but unnecessarily private.

alilleybrinker:
`std::mem::transmute` to get the inner type out of the newtype struct also works, but is use of `unsafe` only to get around a privacy mistake in the API.

Shnatsel:
Fixed by #504

PR #344, opened 2022-12-08
Add support for dependency graph

serbanrobu:
Resolves: https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/14

PR #504, opened 2023-10-26
Write the dependency tree to the SBOM

### LINKED ISSUE #14: Add support for dependency graph introduced in v1.2 of the spec

Shnatsel:
This PR builds on several other PRs and ties everything together. It depends on https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/496, https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/498 and https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/503 - **please review them first!**

Fixes #14 and #409, supersedes #344

I have not tested this in any considerable detail. I would appreciate if you could verify that this parses successfully using the tools you care about.

### DIFF
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -98,7 +98,7 @@ impl SbomGenerator {
                     top_level_dependencies(member, &packages, &resolve)
                 };
 
-            let bom = create_bom(member, &dependencies)?;
+            let bom = create_bom(member, &dependencies, &resolve)?;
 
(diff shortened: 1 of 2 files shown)


lfrancke (approved):
I tested this on a few projects and the result seems good to me.
I didn't find any tooling that could help me verify the results so I did manual spot check and I'm happy.

We can merge it as is.
One idea (but that's all it is, an idea) is whether we want to make it configurable to write the dependency tree information.
It is not required for all use-cases and can bloat the file.
If required we can implement this toggle later though.

Issue #226, opened 2022-08-12
Feature awareness?


jadamcrain:
My use case is building FFI libraries that can distributed as C binary headers, but also embedded in Java Jars and Nuget pacakges.

A couple of questions:

1) Some of my customers want to exclude certain features from builds. Is this tool feature aware?
2) Can the tool exclude build-only dependencies? I've noticed that by default is includes them.
3) Is the purl standardized for Rust somewhere in the spec? What should the purl be for crates not distributed via crates.io?

amy-keibler:
These are all good questions

> 1. Some of my customers want to exclude certain features from builds. Is this tool feature aware?

Not currently. In [this issue](https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/175#issuecomment-1147520539), I outlined a few potential areas I want to investigate. It's one of my highest priority items after we complete the work to integrate the library with the cargo plugin.

> 2. Can the tool exclude build-only dependencies? I've noticed that by default is includes them.

Not currently. This somewhat falls under the first point, but it's probably worth having a separate GitHub issue to track it.

> 3. Is the purl standardized for Rust somewhere in the spec? What should the purl be for crates not distributed via crates.io?

The [purl specification](https://github.com/package-url/purl-spec#purl) does not indicate a required `type` specific to Rust, beyond

> type: the package "type" or package "protocol" such as maven, npm, nuget, gem, pypi, etc. Required.

We could interpret this to mean that `cargo` is the "protocol", which would cover third party repositories. This probably needs its own GitHub issue for tracking.

jadamcrain:
Thanks for these detailed answers!

lfrancke:
We should now exclude build only dependencies in main and #512 and #513 will help with selecting features and targets

Shnatsel:
To clarify: only dev-dependencies are excluded. Build dependencies are included in the SBOM because unlike dev-dependencies they can influence the final binary.

lfrancke:
Ah yes, sorry, my bad.

jadamcrain:
Including build dependencies make sense! Thanks for the update.

lfrancke:
So that should fix points 1 and 2 of your list. I don't think we have tackled purls for non crates things yet anywhere but I might have missed it as well.

Shnatsel:
There is a tracking issue fir pURL from sources other than crates.io: https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/501

[...]

PR #512, opened 2023-10-27
Add feature selection via CLI arguments

### LINKED ISSUE #508: Allow selecting Cargo features: `--no-default-features`, `--all-features`, `--features=...`
Right now `cargo cyclonedx` always outputs the SBOM for the default features.

It must be possible to have the SBOM match the Cargo features actually used for the build, because that influences the dependency tree. Matching the Cargo flags for controlling features, `--no-default-features`, `--all-features`, `--features=...`, would be the most intuitive.

Shnatsel:
Supports `--no-default-features`, `--all-features`,` --features=...` CLI arguments. They are fully compatible with Cargo itself, which is verified by the added tests.

This allows selecting the feature configuration that matches the way you build your binaries, so that the dependency list in the SBOM can actually correspond to the binary.

Fixes #508

### DIFF
--- a/cargo-cyclonedx/src/cli.rs
+++ b/cargo-cyclonedx/src/cli.rs
@@ -109,12 +109,14 @@ impl Args {
             None
         } else {
             let mut feature_list: Vec<String> = Vec::new();
-            // Features can be comma-separated for compatibility with Cargo,
+            // Features can be comma- or space-separated for compatibility with Cargo,
             // but only in command-line arguments.
             for comma_separated_features in &self.features {
                 // Feature names themselves never contain commas.
-                for feature in comma_separated_features.split(',') {
-                    feature_list.push(feature.to_owned());
+                for space_separated_features in comma_separated_features.split(',') {
+                    for feature in space_separated_features.split(' ') {
+                        feature_list.push(feature.to_owned());
+                    }
                 }
             }
 
--- a/cargo-cyclonedx/src/cli.rs
+++ b/cargo-cyclonedx/src/cli.rs
@@ -157,3 +157,43 @@ pub enum ArgsError {
     #[error("Invalid prefix from CLI")]
     CustomPrefixError(#[from] PrefixError),
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn parse_features() {
+        let args = vec!["cyclonedx"];
+        let config = parse_to_config(&args);
+        assert!(config.features.is_none());
+
+        let args = vec!["cyclonedx", "--features=foo"];
+        let config = parse_to_config(&args);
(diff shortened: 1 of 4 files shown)


lfrancke:
I believe this will also fix #226

cargo-cyclonedx/src/cli.rs
```diff
@@ -37,6 +37,20 @@ pub struct Args {
     #[clap(long = "quiet", short = 'q')]
     pub quiet: bool,
 
+    // The feature selection flags are not mutually exclusive in Cargo,
+    // so we keep the same behavior here too.
```
lfrancke, line 41: I only understand this comment after having read all the code and did some research myself.

This code does not belong to `all_features`, it means that all three of these new flags are not mutually exclusive.
If that understanding is correct: I think it's a useful comment to have but it needs clarification.

Example: "The three feature selection flags (all_features, no_default_features, features) are not ..."

cargo-cyclonedx/src/cli.rs
```diff
@@ -88,6 +102,33 @@ impl Args {
...
 
+        let features = if self.all_features == false
+            && self.no_default_features == false
+            && self.features.is_empty()
+        {
+            None
+        } else {
+            let mut feature_list: Vec<String> = Vec::new();
+            // Features can be comma- or space-separated for compatibility with Cargo,
+            // but only in command-line arguments.
```
lfrancke, line 113: What does this second part mean? ("but only in..")

cargo-cyclonedx/src/cli.rs
```diff
@@ -117,3 +159,49 @@ pub enum ArgsError {
     #[error("Invalid prefix from CLI")]
     CustomPrefixError(#[from] PrefixError),
 }
+
+#[cfg(test)]
+mod tests {
```
lfrancke, line 164: Thanks for adding tests!

lfrancke (changes requested):
As usual: Only minor comments :)
Thanks for this PR!

cargo-cyclonedx/src/cli.rs
```diff
@@ -37,6 +37,20 @@ pub struct Args {
     #[clap(long = "quiet", short = 'q')]
     pub quiet: bool,
 
+    // The feature selection flags are not mutually exclusive in Cargo,
+    // so we keep the same behavior here too.
+    /// Activate all available features
+    #[clap(long = "all-features")]
+    pub all_features: bool,
```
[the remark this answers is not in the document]
Shnatsel: I've copied the help messages from Cargo itself here.

I'll update the README before release, because otherwise I'll run into lots of merge conflicts. That text is not neatly split into lines, unlike code.

PR #513, opened 2023-10-27
Allow selecting the target with `--target=<TRIPLE>` or `--all-targets`

### LINKED ISSUE #500: Allow emitting SBOM for a specific platform
The generated SBOM currently includes the dependencies from all platforms it could possibly be compiled for. This is true both before and after the `cargo metadata` conversion.

This is how you end up with various `windows` crates in the output even when running the tool on Linux and Mac, but also `hermit-abi` which you've probably never even heard of.

There should be an option to emit the SBOM for a specific platform. This is trivial to implement using `cargo metadata`.

The part I'm not sure about is what the default should be. Should we emit the SBOM for the current target (e.g. Linux if you're on Linux), or emit all possible dependencies including `hermit-abi` etc?

Shnatsel:
**Depends on  #512, please review that first.**

`cargo cyclonedx` now defaults to the host platform, producing a SBOM for the binary you would get with `cargo build --release`.

"Cross-compilation" of the SBOM is natively supported with `--target=<TRIPLE>`, same flag as for `cargo build`.

The old behavior that includes packages from all targets, even the packages only used on that one funky unikernel, is still available via `--all-targets`.

Fixes #500

### DIFF
--- a/cargo-cyclonedx/src/cli.rs
+++ b/cargo-cyclonedx/src/cli.rs
@@ -52,8 +52,13 @@ pub struct Args {
     #[clap(long = "features", short = 'F')]
     pub features: Vec<String>,
 
-    /// The target to generate the SBOM for, e.g. x86_64-unknown-linux-gnu
-    #[clap(long = "target")]
+    /// The target to generate the SBOM for, or 'all' for all targets.
+    #[clap(
+        long = "target",
+        long_help = "The target to generate the SBOM for, e.g. 'x86_64-unknown-linux-gnu'.
+Use 'all' to include dependencies for all possible targets.
+Defaults to the host target, as printed by 'rustc -vV'"
+    )]
     pub target: Option<String>,
 
     /// Include the dependencies from all possible target platforms in the SBOM
--- a/cargo-cyclonedx/src/cli.rs
+++ b/cargo-cyclonedx/src/cli.rs
@@ -109,12 +109,14 @@ impl Args {
             None
         } else {
             let mut feature_list: Vec<String> = Vec::new();
-            // Features can be comma-separated for compatibility with Cargo,
+            // Features can be comma- or space-separated for compatibility with Cargo,
             // but only in command-line arguments.
             for comma_separated_features in &self.features {
                 // Feature names themselves never contain commas.
-                for feature in comma_separated_features.split(',') {
-                    feature_list.push(feature.to_owned());
+                for space_separated_features in comma_separated_features.split(',') {
+                    for feature in space_separated_features.split(' ') {
+                        feature_list.push(feature.to_owned());
+                    }
                 }
             }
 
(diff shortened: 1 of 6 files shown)


cargo-cyclonedx/src/cli.rs
```diff
@@ -37,6 +38,29 @@ pub struct Args {
...
+    #[clap(long = "no-default-features")]
+    pub no_default_features: bool,
+
+    /// Space or comma separated list of features to activate
+    #[clap(long = "features", short = 'F')]
+    pub features: Vec<String>,
+
+    /// The target to generate the SBOM for, e.g. x86_64-unknown-linux-gnu
+    #[clap(long = "target")]
+    pub target: Option<String>,
```
lfrancke, line 57: These need to be added to the README and the default needs to be documented (host platform) in the readme and in here I'd say

cargo-cyclonedx/src/platform.rs
```diff
@@ -0,0 +1,31 @@
...
+
+pub fn rustc_location() -> OsString {
+    // Honor the environment variable used by Cargo to locate `rustc`:
+    // https://doc.rust-lang.org/cargo/reference/environment-variables.html
+    std::env::var_os("RUSTC").unwrap_or("rustc".into())
+}
+
+/// Returns the default target triple for the rustc we're running
+pub fn rustc_host_target_triple(rustc_path: &OsStr) -> String {
+    // While this feels somewhat insane, this is how `cargo` determines the host platform
```
lfrancke, line 20: :+1: :laughing:

cargo-cyclonedx/src/cli.rs
```diff
@@ -37,6 +38,29 @@ pub struct Args {
...
+
+    /// Include the dependencies from all possible target platforms in the SBOM
+    #[clap(long = "all-targets")]
+    #[clap(conflicts_with = "target")]
+    pub all_targets: bool,
```
lfrancke, line 62: I _think_ I like it better your way but `cargo tree` uses `--target all` for this.

lfrancke (changes requested):
Same as before: README update missing, otherwise looking good.

I tried it (together with the dependency tree PR as well) and it looks good!

Shnatsel:
I'm going to round up the README changes before release. README is not neat split into small lines, so I'll run into lots of merge conflicts if I edit it in every PR.

cargo-cyclonedx/src/cli.rs
Shnatsel: I've changed it to the `cargo tree` way for consistency. I'm not a fan of it, but following a convention is usually preferable.

lfrancke: Fine with me as well. A comment would be useful why we make this decision for later generations to understand.

PR #516, opened 2023-10-28
Fix release name

keponk:
I noticed embarrassingly late that I named the tags wrong. This fixes the wrong naming convention.

Also, replaced the rust installer, the action which was used was throwing warnings, plus I saw the source repo of if has been made public archive. This new one seems to be the more popular active one.


lfrancke (approved):

lfrancke:
Looks good to me but I'll give others a chance to look as well

Shnatsel (approved):

PR #520, opened 2023-10-30
Remove configuration through Cargo.toml

Shnatsel:
This is a fundamentally wrong place for configuration. The tool should only be configurable by the end user, not the developers of the project being audited. See https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/495 for details.

The code could be exhumed from the git history later and reused for a configuration file controlled only by the end user. But I'll spend time adding that only once clear use cases for it emerge, i.e. someone I can talk to and ask for details of their use case requests it.


Shnatsel:
This also requires a README update, but I'll do that later for all PRs at once because there will be too many editing conflicts otherwise.

lfrancke:
Looking good to me but I'm only on my phone today, I'll test this tomorrow.

lfrancke (approved):
Tried, tested, reviewed: Looks good!

Issue #231, opened 2022-08-21
Investigate if custom registries should result in purl's with a different type from "cargo"


amy-keibler:
From #226's

> 3. Is the purl standardized for Rust somewhere in the spec? What should the purl be for crates not distributed via crates.io?

The [purl specification](https://github.com/package-url/purl-spec#purl) does not indicate a required type specific to Rust, beyond

> type: the package "type" or package "protocol" such as maven, npm, nuget, gem, pypi, etc. Required.

Investigate what other CycloneDX tools are doing, particularly if they support private repositories. Rust supports [private registries](https://doc.rust-lang.org/cargo/reference/registries.html) and indicates at a per-dependency level what registry it comes from, so we should be able to access this information if we want to use that as the purl's `type` for a dependency component and the `[package]`'s `publish` list for the package's component

Shnatsel:
There is a dedicated qualifier for this:

> `repository_url` is an extra URL for an alternative, non-default package repository or registry. 

Source: https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#known-qualifiers-keyvalue-pairs

This is what is used in #523

Issue #514, opened 2023-10-28
Reproducible SBOMs


lfrancke:
CycloneDX 1.6 (I believe) allows producing reproducible SBOMs by leaving out timestamps. We should add an option to alow the same.

lfrancke:
We'd love to take a PR for this if you're interested.
We can't promise that someone will get to it soon I'm afraid :(

Shnatsel:
See also  #556 for some relevant discussion

PR #523, opened 2023-10-31
Encode the origin of the package in PURL

### LINKED ISSUE #501: Include information on dependency origin (crates.io, git, custom registry)
Right now `cargo cyclonedx` pretends that all dependencies are released versions that come from crates.io. This is a lie.

Shnatsel:
Fixes #501 and #231

Supersedes #69


cargo-cyclonedx/src/urlencode.rs
```diff
@@ -0,0 +1,85 @@
...
+//! <https://url.spec.whatwg.org/#query-percent-encode-set>
+//!
+//! Note the absence of the `+` character in this set!
+//! It is apparently legal to put a + in there, but the parsers I tried convert it into a space!
+//!
+//! There are only two character sets that escape `+`:
+//! 1. <https://url.spec.whatwg.org/#component-percent-encode-set>
+//!    to be used for "components", but the spec NEVER DEFINES WHAT A COMPONENT IS.
+//! 2. <https://url.spec.whatwg.org/#application-x-www-form-urlencoded-percent-encode-set>
+//!    to be used for form submission, so not our case?
```
kylewillmon, line 30: It's a mistake to believe that this is only used for form submission. This same exact encoding is used to set the `query` string of a URL (because that's where form submission data goes when the method is `GET` anyways).

The spec even states that this set will be used [for anything in the `URLSearchParams` class][0], which is used for generating query strings (or body data in the case of `POST` form submission).

[0]: https://url.spec.whatwg.org/#ref-for-application-x-www-form-urlencoded-percent-encode-set%E2%91%A3

That is why, for example, the [whatwg-url][1] package (which is [synchronized][2] with the WHATWG URL spec), encodes these characters when set using the `searchParam` attribute. For example:

```js
import { URL } from 'whatwg-url';
const x = new URL('pkg:generic/bitwarderl');
x.searchParams.set('vcs_url', 'vcs_url=git+https://github.com/rustsec/rustsec.git@6f82509cb43a')
console.log(x.toString());
```

Outputs:

```
pkg:generic/bitwarderl?vcs_url=vcs_url%3Dgit%2Bhttps%3A%2F%2Fgithub.com%2Frustsec%2Frustsec.git%406f82509cb43a
```

[1]: https://www.npmjs.com/package/whatwg-url
[2]: https://github.com/whatwg/url#implementations

It is fine to use characters such as `+` unencoded in the query string, but **not** if you are parsing that query string as search parameters (as PURL does).

Shnatsel: Even that cannot be used to produce the official example of `?vcs_url=git%2Bhttps://git.fsfe.org/dxtr/bitwarderl%40cc55108da32`, because the `application-x-www-form-urlencoded` set encodes `:` and `/` but they are left unencoded in the official example.

kylewillmon: I suppose it would be a question to the PURL maintainers to see if they actually intend qualifiers to be interoperable with `URLSearchParams`. If so, mandatory percent encoding of `+` should be mentioned in the spec...

lfrancke: @pombredanne is this something you can maybe weigh in on?

cargo-cyclonedx/src/generator.rs
```diff
@@ -169,6 +171,42 @@ impl SbomGenerator {
         component
     }
 
+    fn get_purl(package: &Package) -> Result<Purl, purl::PackageError> {
```
lfrancke, line 174: I think this method needs unit tests

cargo-cyclonedx/src/generator.rs
```diff
@@ -169,6 +171,42 @@ impl SbomGenerator {
...
+                        log::error!("No '+' separator found in source field from `cargo metadata`")
+                    }
+                }
+            }
+        } else {
+            // source is None for packages from the local filesystem.
+            // The manifest path ends with a `Cargo.toml`, so the package directory is its parent
+            let package_dir = package.manifest_path.parent().unwrap();
+            // url-encode the path to the package manifest to make it a valid URL
+            let manifest_url = format!("file://{}", urlencode(package_dir.as_str()));
```
lfrancke, line 201: I don't like this at all. This encodes the absolute path for workspace projects (like this).
It makes the SBOM not reproducible.

I do understand the reasoning but I would hope that there is a better way.
I'll ask on Slack.

lfrancke (commented):
I have not looked at urlencode.rs yet

cargo-cyclonedx/src/generator.rs
Shnatsel: I'm reasonably confident that it is correct given my assumptions of how `cargo metadata` works. What it needs is integration tests to test it against real `cargo metadata` rather than synthetic data I made up, which is still limited by my assumptions.

cargo-cyclonedx/src/generator.rs
Shnatsel: Summarizing from the discussion on Slack:

`pkg:cargo/cyclonedx-bom@0.4.1?download_url=file://../cyclonedx-sbom/` (a relative url instead of an absolute one) would at least solve the reproducibility issue (for PURL; there are other sources of non-determinism).

I think the only way to deal with such cases gracefully is to bypass local checkouts completely. Instead point the tool at the package URL, have it download it by itself so that it knows where it came from (default registry, custom registry, VCS URL) and include that verified source into the PURL and metadata. But that's out of scope of this PR.

Shnatsel:
Turning into a draft until the reproducibility issues are addressed.

Shnatsel:
I've implemented relative paths for packages within the workspace to allow for more reproducible builds (at least in PURLs, there are other aspects of the SBOM that are not reproducible).

This needs even more tests but other than that it's ready to go, PTAL.

lfrancke:
For full reproducability we have [another issue](https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/514)

cargo-cyclonedx/src/purl.rs
```diff
@@ -0,0 +1,231 @@
...
+    const ROOT_PACKAGE_JSON: &str = include_str!("../tests/fixtures/root_package.json");
+    const WORKSPACE_PACKAGE_JSON: &str = include_str!("../tests/fixtures/workspace_package.json");
+
+    #[test]
+    fn crates_io_purl() {
+        let crates_io_package: Package = serde_json::from_str(CRATES_IO_PACKAGE_JSON).unwrap();
+        let purl = get_purl(
+            &crates_io_package,
+            &crates_io_package,
+            &Utf8Path::new("/foo/bar"),
```
lfrancke, line 106:
```suggestion
            Utf8Path::new("/foo/bar"),
```

cargo-cyclonedx/src/purl.rs
```diff
@@ -0,0 +1,231 @@
...
+        assert_eq!(parsed_purl.version(), Some("1.1.2"));
+        assert!(parsed_purl.qualifiers().is_empty());
+        assert!(parsed_purl.subpath().is_none());
+        assert!(parsed_purl.namespace().is_none());
+    }
+
+    #[test]
+    fn git_purl() {
+        let git_package: Package = serde_json::from_str(GIT_PACKAGE_JSON).unwrap();
+        let purl = get_purl(&git_package, &git_package, &Utf8Path::new("/foo/bar"), None).unwrap();
```
lfrancke, line 122:
```suggestion
        let purl = get_purl(&git_package, &git_package, Utf8Path::new("/foo/bar"), None).unwrap();
```

cargo-cyclonedx/src/purl.rs
```diff
@@ -0,0 +1,231 @@
...
+        assert!(parsed_purl.namespace().is_none());
+    }
+
+    #[test]
+    fn toplevel_package_purl() {
+        let root_package: Package = serde_json::from_str(ROOT_PACKAGE_JSON).unwrap();
+        let purl = get_purl(
+            &root_package,
+            &root_package,
+            &Utf8Path::new("/home/shnatsel/Code/cargo-cyclonedx/"),
```
lfrancke, line 141:
```suggestion
            Utf8Path::new("/home/shnatsel/Code/cargo-cyclonedx/"),
```

cargo-cyclonedx/src/purl.rs
```diff
@@ -0,0 +1,231 @@
...
+        assert!(parsed_purl.namespace().is_none());
+    }
+
+    #[test]
+    fn toplevel_package_with_subpath() {
+        let root_package: Package = serde_json::from_str(ROOT_PACKAGE_JSON).unwrap();
+        let purl = get_purl(
+            &root_package,
+            &root_package,
+            &Utf8Path::new("/home/shnatsel/Code/cargo-cyclonedx/"),
```
lfrancke, line 164:
```suggestion
            Utf8Path::new("/home/shnatsel/Code/cargo-cyclonedx/"),
```

cargo-cyclonedx/src/purl.rs
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    #[test]
+    fn workspace_package() {
+        let root_package: Package = serde_json::from_str(ROOT_PACKAGE_JSON).unwrap();
+        let workspace_package: Package = serde_json::from_str(WORKSPACE_PACKAGE_JSON).unwrap();
+        let purl = get_purl(
+            &workspace_package,
+            &root_package,
+            &Utf8Path::new("/home/shnatsel/Code/cargo-cyclonedx/"),
```
lfrancke, line 188:
```suggestion
            Utf8Path::new("/home/shnatsel/Code/cargo-cyclonedx/"),
```

cargo-cyclonedx/src/purl.rs
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    #[test]
+    fn local_package() {
+        let root_package: Package = serde_json::from_str(ROOT_PACKAGE_JSON).unwrap();
+        let workspace_package: Package = serde_json::from_str(WORKSPACE_PACKAGE_JSON).unwrap();
+        let purl = get_purl(
+            &workspace_package,
+            &root_package,
+            &Utf8Path::new("/foo/bar/"),
```
lfrancke, line 212:
```suggestion
            Utf8Path::new("/foo/bar/"),
```

lfrancke (changes requested):
The clippy warnings are only in test code but still. Otherwise looks good. Thanks for the tests and the attention to detail!

lfrancke (approved):

lfrancke:
This means #69 can be closed, right?

PR #525, opened 2023-10-31
Fix dev-dependencies being included in the dependency tree

Shnatsel:
1. Make validation failure of the generated SBOM panic when debug assertions are enabled (as opposed to having it only log the issue with `log_level=debug`)
2. Only run potentially quadratic validation when debug assertions are enabled
3. Fix the issue that caused dev-dependencies to be included in the SBOM and validation to fail


lfrancke (approved):
LGTM!

Thanks for fixing this so fast

PR #526, opened 2023-10-31
Default to listing all dependencies, not just top-level ones

Shnatsel:
1. Default to listing all dependencies, not just top-level ones. Now that the platform filtering is implemented, this will match the components that go into the final binary.
2. Make `--all` and `--top-level` flags incompatible with each other, because passing both makes no sense

### DIFF
--- a/cargo-cyclonedx/src/config.rs
+++ b/cargo-cyclonedx/src/config.rs
@@ -1,5 +1,6 @@
 use serde::Deserialize;
 use std::collections::HashSet;
+use std::default;
 use std::str::FromStr;
 use thiserror::Error;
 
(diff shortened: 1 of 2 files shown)


cargo-cyclonedx/src/config.rs
```diff
@@ -1,5 +1,6 @@
 use serde::Deserialize;
 use std::collections::HashSet;
+use std::default;
```
lfrancke, line 3: Rust complains about this being unused

lfrancke (changes requested):

Shnatsel:
Fixed, PTAL.

lfrancke (approved):

PR #527, opened 2023-10-31
Offload toplevel dependency selection to `cargo metadata`

Shnatsel:
Use `cargo metadata --no-deps` when listing top-level dependencies only. 

(Instead of rolling our own custom logic for this which is invariably going to be buggy _somehow._)

### DIFF
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -73,19 +73,25 @@ impl SbomGenerator {
     ) -> Result<Vec<GeneratedSbom>, GeneratorError> {
         log::trace!("Processing the workspace {}", meta.workspace_root);
         let members: Vec<PackageId> = meta.workspace_members;
-        let packages = index_packages(meta.packages);
-        let resolve = index_resolve(meta.resolve.unwrap().nodes);
+        let mut packages = index_packages(meta.packages);
+        // When listing only top-level dependencies, `cargo metadata` will not output the dependency tree
+        let mut resolve = meta.resolve.map(|r| index_resolve(r.nodes));
 
         let mut result = Vec::with_capacity(members.len());
         for member in members.iter() {
             log::trace!("Processing the package {}", member);
 
-            let (dependencies, pruned_resolve) =
-                prune_dev_dependencies(member, &packages, &resolve);
+            if let Some(res) = resolve {
+                let (pruned_packages, pruned_resolve) =
+                    prune_dev_dependencies(member, &packages, &res);
+                packages = pruned_packages;
+                resolve = Some(pruned_resolve);
+            }
+
             let generator = SbomGenerator {
                 config: config.clone(),
             };
-            let bom = generator.create_bom(member, &dependencies, &pruned_resolve)?;
+            let bom = generator.create_bom(member, &packages, &resolve)?;
 
             if cfg!(debug_assertions) {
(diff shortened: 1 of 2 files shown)


Shnatsel:
I would like to also rename this flag from `--top-level` to `--no-deps` for consistency with `cargo metadata` and it generally being more descriptive. What do you think?

lfrancke:
> I would like to also rename this flag from `--top-level` to `--no-deps` for consistency with `cargo metadata` and it generally being more descriptive. What do you think?

I see what you're getting at but both the CISA and the BSI call it "top-level" dependencies.
So from someone wanting to create an SBOM "top-level" is the better term.
For a Rust person "no-deps" might be better. We could allow both as aliases, if we only keep one I'm in favor of "top-level"

Shnatsel:
Fair enough, I'll keep the top-level terminology then. Thanks!

lfrancke:
I just ran this and the results seem wrong.
This seems to only give me other workspace dependencies.
As in cargo-cyclonedx only has cyclonedx-bom as a component and cyclonedx-bom lists cargo-cyclonedx.

Shnatsel:
Ah, it seems `cargo metadata --no-deps` does something different from what is desired here. I'll close this, then. Thanks for testing!

lfrancke:
You might want to double check what I did.
It's late and I might have missed something. The actual JSON output from cargo metadata --no-deps _seems_ like what we want? Maybe it's just the processing or the `cargo-metadata` crate doing something wrong?

Shnatsel:
Could you point me to the spec that explains what "top level" is supposed to be, exactly? I can do my own research from there.

lfrancke:
https://www.ntia.doc.gov/files/ntia/publications/sbom_minimum_elements_report.pdf

> Depth
> An SBOM should contain all primary (top level) components, with all their transitive
dependencies listed. At a minimum, all top-level dependencies must be listed with enough
detail to seek out the transitive dependencies recursively.
> Going further into the graph will provide more information. As organizations begin SBOM,
depth beyond the primary components may not be easily available due to existing
requirements with subcomponent suppliers. Eventual adoption of SBOM processes will
enable access to additional depth through deeper levels of transparency at the subcomponent
level. It should be noted that some use cases require complete or mostly complete graphs,
such as the ability to “prove the negative” that a given component is not on an organization’s
network.
> An SBOM consumer can specify depth by number of transitive steps. Alternatively, they
could specify depth in operational terms. This could include software attributes, such as “all
non-open source software,” or all components of a certain function or complexity.
Organizations can also incentivize greater reporting depth and completeness by offering
different requirements on reporting and remediation of vulnerabilities in components
enumerated in the SBOM versus those that are not. Such specifications are outside the scope
of these minimum elements.

https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/TechnischeRichtlinien/TR03183/BSI-TR-03183-2.pdf?__blob=publicationFile&v=3 (German only, english version in the works)

> Top-Level-SBOM
> Die SBOM beinhaltet neben den Informationen zur Primärkomponente Informationen zu allen
Komponenten, die direkt (unmittelbar) durch die Primärkomponente eingebunden sind. Die SBOM wird
auch als „SBOM der obersten Ebene“ bezeichnet

Shnatsel:
Ah, indeed `cargo metadata --no-deps` is not what we want. We'll have to keep the custom filtering. Thanks!

PR #529, opened 2023-11-01
Record the rustc target platform into the SBOM

### LINKED ISSUE #528: Record the target platform in the SBOM
The target for which the SBOM is generated should be recorded in it.

It would be nice to also append it to the end of the SBOM filename, to make it very clear that the SBOM is generated for a specific platform (with an option to disable this).

Shnatsel:
Fixes #528

### DIFF
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -317,10 +317,10 @@ impl SbomGenerator {
         metadata.tools = Some(Tools(vec![tool]));
 
         let properties = match self.config.target.as_ref().unwrap() {
-            SingleTarget(target) => vec![Property::new("rustcTarget", target)],
+            SingleTarget(target) => vec![Property::new("rustc:target", target)],
             AllTargets => all_known_targets()
                 .into_iter()
-                .map(|target| Property::new("rustcTarget", &target))
+                .map(|target| Property::new("rustc:target", &target))
                 .collect(),
         };
         metadata.properties = Some(Properties(properties));
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -317,7 +317,7 @@ impl SbomGenerator {
         metadata.tools = Some(Tools(vec![tool]));
 
         let properties = match self.config.target.as_ref().unwrap() {
-            SingleTarget(target) => vec![Property::new("rustcTarget", &target)],
+            SingleTarget(target) => vec![Property::new("rustcTarget", target)],
             AllTargets => all_known_targets()
                 .into_iter()
                 .map(|target| Property::new("rustcTarget", &target))
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -18,6 +18,7 @@
 use crate::config::Pattern;
 use crate::config::Prefix;
 use crate::config::SbomConfig;
+use crate::config::Target::SingleTarget;
 use crate::config::{IncludedDependencies, ParseMode};
 use crate::format::Format;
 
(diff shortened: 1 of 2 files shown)


cargo-cyclonedx/src/generator.rs
```diff
@@ -311,6 +316,15 @@ impl SbomGenerator {
 
         metadata.tools = Some(Tools(vec![tool]));
 
+        let properties = match self.config.target.as_ref().unwrap() {
+            SingleTarget(target) => vec![Property::new("rustcTarget", &target)],
+            AllTargets => all_known_targets()
+                .into_iter()
+                .map(|target| Property::new("rustcTarget", &target))
```
lfrancke, line 323: I like the idea in general. Have you checked if there are any plans/standards/best practices for what this property could be called? Other tools must have the same idea/concept in general (e.g. Go/C/C++)

lfrancke (commented):
(didn't look at the code yet, just one question)

cargo-cyclonedx/src/generator.rs
Shnatsel: I have not really checked for other uses of `properties`, just skimmed the standardized fields of `metadata`. 

This encodes Rust target triples specifically, e.g. `x86_64-unknown-linux-gnu`, so this is not going to be interoperable with anything else. Hence the use of `properties` instead.

cargo-cyclonedx/src/generator.rs
```diff
@@ -311,6 +316,15 @@ impl SbomGenerator {
 
         metadata.tools = Some(Tools(vec![tool]));
 
+        let properties = match self.config.target.as_ref().unwrap() {
+            SingleTarget(target) => vec![Property::new("rustcTarget", &target)],
```
lfrancke, line 320:
```suggestion
            SingleTarget(target) => vec![Property::new("rustcTarget", target)],
```

lfrancke: clippy lint

cargo-cyclonedx/src/generator.rs
lfrancke: Understood. In that case, does it make sense to use a namespaced property instead?

https://github.com/CycloneDX/cyclonedx-property-taxonomy
https://github.com/CycloneDX/cyclonedx-property-taxonomy/blob/main/cdx.md

lfrancke (commented):
Tried, tested, works etc.
Happy to approve in principle but I think I'd like to talk about the name of the property a bit first.
This is the first PR (I believe) to introduce a custom property so it'll be a bit more involved. The next one will be easier.

cargo-cyclonedx/src/generator.rs
lfrancke: Found the property for Go: https://github.com/CycloneDX/cyclonedx-property-taxonomy/blob/main/cdx/gomod.md

Shnatsel: Yes, it does. Good point.

lfrancke:
I think this is basically good to go, we just need to wait for the result of https://github.com/CycloneDX/cyclonedx-property-taxonomy/issues/75

Shnatsel:
That property needs a fair bit of design work. We need to specify where it is legal for it to appear: on metadata only, or does it also appear on a `component` to specify what target is it for? When recording dependencies for all platforms, do we have a boolean flag to indicate it, or do we list all platforms considered? If the latter, does it make sense to make that property an array or a set, instead of specifying it multiple times?

Shnatsel:
I've opened https://github.com/CycloneDX/cyclonedx-property-taxonomy/pull/78/ upstream. Once that's merged, this PR will need to be reworked to match that schema.

Shnatsel:
Superseded by #762

PR #530, opened 2023-11-01
`cyclonedx-bom`: Don't break on newer compilers

Shnatsel:
Remove ![deny(warnings)] so that cyclonedx-bom doesn't fail to build in production just because a newer compiler added a new lint

### DIFF
--- a/cyclonedx-bom/src/lib.rs
+++ b/cyclonedx-bom/src/lib.rs
@@ -16,9 +16,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#![deny(clippy::all)]
-#![deny(warnings)]
-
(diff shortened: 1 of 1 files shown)


lfrancke (approved):

Shnatsel:
Breaking builds on CI - yes.

Breaking builds for other projects who depend on this crate transitively - bad idea.

Shnatsel:
Making warnings fail CI is also a questionable practice. If I didn't have to suppress a warning to get a PR merged, the bug I had to fix in https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/525 would not have happened because there would be a warning about an unused variable.

lfrancke:
I believe we will have to agree to disagree on this one.

PR #534, opened 2023-11-03
Also record PURL when recording multiple binaries

Shnatsel:
Depends on #523 and #533, please review them first

Adds recording PURL for the binary targets of the toplevel package on top of #533


Shnatsel:
Now that support for multiple binaries is merged, there is no reason to keep this in a separate PR. I'll continue working on this in https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/523 which already has some review comments.

PR #535, opened 2023-11-03
Include the target platform in the SBOM filename

Shnatsel:
And provide a `--no-target-suffix` flag to let people opt out.

### DIFF
--- a/cargo-cyclonedx/src/cli.rs
+++ b/cargo-cyclonedx/src/cli.rs
@@ -91,7 +91,8 @@ Defaults to the host target, as printed by 'rustc -vV'"
     #[clap(
         name = "output-prefix",
         long = "output-prefix",
-        value_name = "FILENAME_PREFIX"
+        value_name = "FILENAME_PREFIX",
+        conflicts_with = "output-pattern"
     )]
     pub output_prefix: Option<String>,
 
--- a/cargo-cyclonedx/src/cli.rs
+++ b/cargo-cyclonedx/src/cli.rs
@@ -63,9 +63,9 @@ Defaults to the host target, as printed by 'rustc -vV'"
     )]
     pub target: Option<String>,
 
-    /// Do not include the target platform of the BOM in the filename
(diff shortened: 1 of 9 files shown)


cargo-cyclonedx/src/cli.rs
```diff
@@ -63,6 +63,10 @@ Defaults to the host target, as printed by 'rustc -vV'"
     )]
     pub target: Option<String>,
 
+    /// Do not include the target platform of the BOM in the filename
+    #[clap(long = "no-target-suffix")]
+    pub no_target_suffix: bool,
```
lfrancke, line 68: Personally I think this should be off by default.

cargo-cyclonedx/src/cli.rs
```diff
@@ -154,21 +158,19 @@ impl Args {
...
-                prefix,
-            }),
-            (_, _) => None,
-        };
+        let output_options =
+            if cdx_extension.is_none() && prefix.is_none() && !self.no_target_suffix {
+                None
+            } else {
+                Some(OutputOptions {
+                    cdx_extension: cdx_extension.unwrap_or_default(),
```
lfrancke, line 166: This comment doesn't concern this line in particular but in https://cyclonedx.org/specification/overview/#recognized-file-patterns it is outlined how a valid CycloneDX file is called and this PR violates that.

If it's not `bom.xml` anymore, it should be `bom_x86_64-unknown-linux-gnu.cdx.xml`

lfrancke (changes requested):

lfrancke:
To summarize my review:

- I see that people might want this, I don't think I'll need it but I'm not opposed
- I'd like to make this opt-in rather than opt-out
- We should stay compatible with the recognized file patterns which would imply to use the cdx suffix if target is included

Shnatsel:
I believe I have addressed the review feedback, PTAL.

lfrancke (approved):

PR #539, opened 2023-11-08
Update list of authors

Shnatsel:
Also gets encoded in the generated SBOM

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ members = [
 ]
 
 [workspace.package]
-authors = [ "Steve Springett <steve.springett@owasp.org>" ]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -6,7 +6,7 @@ members = [
 ]
 
 [workspace.package]
-authors = [ "Steve Springett <steve.springett@owasp.org>" ]
+authors = [ "Steve Springett <steve.springett@owasp.org>", "Amy Keibler <3483663+amy-keibler@users.noreply.github.com>", "Sergey \"Shnatsel\" Davidoff <shnatsel@gmail.com>"]
```
lfrancke, line 9: Not sure if we want to put Amy in here without her consent.
I don't have any strong opinions. @amy-keibler ping, just in case

lfrancke (approved):

tshepang:
I would just remove the field, since it won't be fun to maintain, and multiple authors have worked on the project.

PR #543, opened 2023-11-12
Update README

Shnatsel:
The list of options is presented as the auto-generated help text because it is the only way to reasonably keep it in sync with the code. Everything else drifts and confuses users eventually.

### DIFF
--- a/cargo-cyclonedx/src/cli.rs
+++ b/cargo-cyclonedx/src/cli.rs
@@ -96,7 +96,7 @@ Defaults to the host target, as printed by 'rustc -vV'"
     )]
     pub output_prefix: Option<String>,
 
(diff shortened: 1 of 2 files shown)


cargo-cyclonedx/src/cli.rs
```diff
@@ -96,7 +96,7 @@ Defaults to the host target, as printed by 'rustc -vV'"
     )]
     pub output_prefix: Option<String>,
 
-    /// Reject the deprecated '/' separator for licenses, treating 'MIT/Apache-2.0' as 'MIT OR Apache-2.0'
+    /// Reject the deprecated '/' separator for licenses, treating 'MIT/Apache-2.0' as an error
```
lfrancke, line 99: Good catch!

lfrancke (approved):

PR #550, opened 2023-11-13
refactor: implement Display, and AsRef<str>

ctron:
Implementing `Display` allows one to use things like `format!`, and implements `ToString` automatically.

Implementing `AsRef<str>` allows to get an `&str` from the value, without needing to clone, or using a `Cow<str>` when working with other code which does offer `&str` (like in a match).


cyclonedx-bom/src/external_models/normalized_string.rs
```diff
@@ -45,9 +46,15 @@ impl NormalizedString {
     }
 }
 
-impl ToString for NormalizedString {
-    fn to_string(&self) -> String {
-        self.0.clone()
```
Shnatsel, line 50: Removing `ToString` would be a semver break. Are you sure this is justified?

cyclonedx-bom/src/external_models/normalized_string.rs
```diff
@@ -45,9 +46,15 @@ impl NormalizedString {
...
+impl AsRef<str> for NormalizedString {
```
Shnatsel, line 49: Is there a good reason to implement `AsRef` instead of [`Deref`](https://doc.rust-lang.org/stable/std/ops/trait.Deref.html)?

cyclonedx-bom/src/external_models/normalized_string.rs
Shnatsel: Nevermind, an implementation of it is [provided](https://doc.rust-lang.org/stable/std/string/trait.ToString.html) by implementing `Display`.

cyclonedx-bom/src/external_models/normalized_string.rs
ctron: No. Maybe we should have both. `String` also implements `AsRef<str>`, so you can have `fn foo(s: impl AsRef<str>)`.

ctron:
Added `Deref<Target=str>` in addition. It could also be `Deref<Target=String>`?

Shnatsel (approved):

PR #553, opened 2023-11-14
Improve logging for kinds that are not binary or library

lfrancke:


cargo-cyclonedx/src/generator.rs
```diff
@@ -182,60 +182,61 @@ impl SbomGenerator {
...
-                    .parent()
-                    .expect("manifest_path in `cargo metadata` output is not a file!");
-                if let Ok(relative_path) = tgt.src_path.strip_prefix(package_dir) {
-                    subcomponent.purl =
-                        get_purl(package, package, &self.workspace_root, Some(relative_path)).ok();
-                } else {
+            // Classification
+            let cdx_type = match (tgt.is_bin(), tgt.is_lib()) {
+                (true, _) => Classification::Application,
+                (_, true) => Classification::Library,
```
Shnatsel, line 188: This could be more explicit by making it `(false, true)` but I don't have a strong preference

cargo-cyclonedx/src/generator.rs
```diff
@@ -182,60 +182,61 @@ impl SbomGenerator {
...
-                // So the information we encode here is sufficient to idenfity the file in git too.
-                let package_dir = package
-                    .manifest_path
-                    .parent()
-                    .expect("manifest_path in `cargo metadata` output is not a file!");
-                if let Ok(relative_path) = tgt.src_path.strip_prefix(package_dir) {
-                    subcomponent.purl =
-                        get_purl(package, package, &self.workspace_root, Some(relative_path)).ok();
-                } else {
+            // Classification
```
Shnatsel, line 185: This PR seems to have dropped an important comment:

```
            // Ignore tests, benches, examples and build scripts.
            // They are not part of the final build artifacts, which is what we are after.
```

I'd like to have it re-added.

lfrancke: Done in https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/553/commits/5127338ae9472a0260969654ab41c3c470d13c3b

cargo-cyclonedx/src/generator.rs
lfrancke: Done in https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/553/commits/5127338ae9472a0260969654ab41c3c470d13c3b

Shnatsel (approved):

cargo-cyclonedx/src/generator.rs
```diff
@@ -182,60 +182,63 @@ impl SbomGenerator {
         let mut subcomponents: Vec<Component> = Vec::new();
         let mut subcomp_count: u32 = 0;
         for tgt in &package.targets {
+            // Classification
             // Ignore tests, benches, examples and build scripts.
             // They are not part of the final build artifacts, which is what we are after.
-            if !(tgt.is_bench() || tgt.is_example() || tgt.is_test() || tgt.is_custom_build()) {
```
Shnatsel, line 187: I missed this during the review, but removing this line was wrong. 

We do need to filter out these things, and a crate will have more than one type at a time.

I'll have to revert this.

lfrancke: I don't think that's correct.
I do check for that by checking for the reverse.
I only "include" lib and bin and continue otherwise.

lfrancke:
```
let cdx_type = match (tgt.is_bin(), tgt.is_lib()) {
                (true, false) => Classification::Application,
                (false, true) => Classification::Library,
                _ => {
                    log::warn!(
                        "Target {} is neither a binary nor a library! Kinds: {}",
                        tgt.name,
                        tgt.kind.join(", ")
                    );

                    continue;
                } // Skip if neither a binary nor a library
            };
```

Anything that is not `is_bin` or `is_lib` will be logged and skipped, just as before.
This should be more resilient should new kinds be added later.

Am I missing something?

Shnatsel: Yes, you are. It's an array of kinds, not a single specific kind. So `.is_example()` and `.is_bin()` are often true the same time.

Issue #552, opened 2023-11-14
Add support for `cargo binstall`


liamwh:
As described [here](https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md), so that `cargo-binstall` doesn't attempt to install cargo-cyclonedx from source:
![image](https://github.com/CycloneDX/cyclonedx-rust-cargo/assets/23312359/134ee46f-143c-49a3-b24e-ba3a5565a584)

lfrancke:
Are you willing or interested in providing a PR for this?

tshepang:
cargo-dist would make providing such extra easy

Shnatsel:
I'd be happy to have a PR for it! We should also update the repo's README listing it as a possible installation option.

It'd be great to get a release workflow with `cross` going for building aarch64 binaries as well, but that's a separate issue (cc @keponk )

ashleygwilliams:
put a PR up- cargo-dist can add the release instructions to the release body- but i'm also happy to add the installers to the README once we decide which ones you want to support.

liamwh:
I've had a really great experience automating releases using [release-plz](https://github.com/MarcoIeni/release-plz). I believe it's even simpler than `cargo-dist`, although I don't have any firsthand experience with `cargo-dist` myself. However I can recommend using release-plz, I found it extremely easy to use and have no complaints with it. Maybe it's helpful here?

ashleygwilliams:
hey @liamwh - release plz is great and i think it can be a good tool for generating changelogs, but it doesn't have functionality for building standalone releases or generating installers. many folks use release-plz with cargo-dist for changelog generation. the main thing that cargo-dist is doing here is building binaries and then archiving those binaries for a variety of platforms, then generating installers for them.

ashleygwilliams:
i believe that this was solved by #727 !

PR #559, opened 2023-11-15
feat(dist): add cargo-dist

ashleygwilliams:
addresses #552 - marked as draft as i made some calls about what targets and installers you'd like to support. happy to talk through those and update.

you can run `cargo dist plan` locally to see what this workflow will generate. i've also put the output here:

```
announcing v0.4.0
  cargo-cyclonedx 0.4.0
    cargo-cyclonedx-installer.sh
    cargo-cyclonedx-installer.ps1
    cargo-cyclonedx-aarch64-apple-darwin.tar.xz
      [bin] cargo-cyclonedx
      [misc] CHANGELOG.md, LICENSE, README.md
      [checksum] cargo-cyclonedx-aarch64-apple-darwin.tar.xz.sha256
    cargo-cyclonedx-x86_64-apple-darwin.tar.xz
      [bin] cargo-cyclonedx
      [misc] CHANGELOG.md, LICENSE, README.md
      [checksum] cargo-cyclonedx-x86_64-apple-darwin.tar.xz.sha256
    cargo-cyclonedx-x86_64-pc-windows-msvc.zip
      [bin] cargo-cyclonedx.exe
      [misc] CHANGELOG.md, LICENSE, README.md
      [checksum] cargo-cyclonedx-x86_64-pc-windows-msvc.zip.sha256
    cargo-cyclonedx-x86_64-unknown-linux-gnu.tar.xz
      [bin] cargo-cyclonedx
      [misc] CHANGELOG.md, LICENSE, README.md
      [checksum] cargo-cyclonedx-x86_64-unknown-linux-gnu.tar.xz.sha256
    cargo-cyclonedx-x86_64-unknown-linux-musl.tar.xz
      [bin] cargo-cyclonedx
      [misc] CHANGELOG.md, LICENSE, README.md
      [checksum] cargo-cyclonedx-x86_64-unknown-linux-musl.tar.xz.sha256
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,266 @@
+# Copyright 2022-2023, axodotdev
+# SPDX-License-Identifier: MIT or Apache-2.0
+#
+# CI that:
+#
+# * checks for a Git Tag that looks like a release
+# * builds artifacts with cargo-dist (archives, installers, hashes)
+# * uploads those artifacts to temporary workflow zip
+# * on success, uploads the artifacts to a Github Release
+#
+# Note that the Github Release will be created with a generated
+# title/body based on your changelogs.
+
+name: Release
+
+permissions:
+  contents: write
+
+# This task will run whenever you push a git tag that looks like a version
+# like "1.0.0", "v0.1.0-prerelease.1", "my-app/0.1.0", "releases/v1.0.0", etc.
+# Various formats will be parsed into a VERSION and an optional PACKAGE_NAME, where
+# PACKAGE_NAME must be the name of a Cargo package in your workspace, and VERSION
+# must be a Cargo-style SemVer Version (must have at least major.minor.patch).
+#
+# If PACKAGE_NAME is specified, then the announcement will be for that
+# package (erroring out if it doesn't have the given version or isn't cargo-dist-able).
+#
+# If PACKAGE_NAME isn't specified, then the announcement will be for all
+# (cargo-dist-able) packages in the workspace with that version (this mode is
+# intended for workspaces with only one dist-able package, or with all dist-able
+# packages versioned/released in lockstep).
+#
+# If you push multiple tags at once, separate instances of this workflow will
+# spin up, creating an independent announcement for each one. However Github
+# will hard limit this to 3 tags per commit, as it will assume more tags is a
+# mistake.
+#
+# If there's a prerelease-style suffix to the version, then the release(s)
+# will be marked as a prerelease.
+on:
+  push:
+    tags:
+      - '**[0-9]+.[0-9]+.[0-9]+*'
+  pull_request:
+
+jobs:
+  # Run 'cargo dist plan' (or host) to determine what tasks we need to do
+  plan:
+    runs-on: ubuntu-latest
+    outputs:
+      val: ${{ steps.plan.outputs.manifest }}
+      tag: ${{ !github.event.pull_request && github.ref_name || '' }}
+      tag-flag: ${{ !github.event.pull_request && format('--tag={0}', github.ref_name) || '' }}
+      publishing: ${{ !github.event.pull_request }}
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: Install cargo-dist
+        # we specify bash to get pipefail; it guards against the `curl` command
+        # failing. otherwise `sh` won't catch that `curl` returned non-0
+        shell: bash
+        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/cargo-dist/releases/download/v0.11.1/cargo-dist-installer.sh | sh"
+      # sure would be cool if github gave us proper conditionals...
+      # so here's a doubly-nested ternary-via-truthiness to try to provide the best possible
+      # functionality based on whether this is a pull_request, and whether it's from a fork.
+      # (PRs run on the *source* but secrets are usually on the *target* -- that's *good*
+      # but also really annoying to build CI around when it needs secrets to work right.)
+      - id: plan
+        run: |
+          cargo dist ${{ (!github.event.pull_request && format('host --steps=create --tag={0}', github.ref_name)) || 'plan' }} --output-format=json > plan-dist-manifest.json
+          echo "cargo dist ran successfully"
+          cat plan-dist-manifest.json
+          echo "manifest=$(jq -c "." plan-dist-manifest.json)" >> "$GITHUB_OUTPUT"
+      - name: "Upload dist-manifest.json"
+        uses: actions/upload-artifact@v4
+        with:
+          name: artifacts-plan-dist-manifest
+          path: plan-dist-manifest.json
+
+  # Build and packages all the platform-specific things
+  build-local-artifacts:
+    name: build-local-artifacts (${{ join(matrix.targets, ', ') }})
+    # Let the initial task tell us to not run (currently very blunt)
+    needs:
+      - plan
+    if: ${{ fromJson(needs.plan.outputs.val).ci.github.artifacts_matrix.include != null && (needs.plan.outputs.publishing == 'true' || fromJson(needs.plan.outputs.val).ci.github.pr_run_mode == 'upload') }}
+    strategy:
+      fail-fast: false
+      # Target platforms/runners are computed by cargo-dist in create-release.
+      # Each member of the matrix has the following arguments:
+      #
+      # - runner: the github runner
+      # - dist-args: cli flags to pass to cargo dist
+      # - install-dist: expression to run to install cargo-dist on the runner
+      #
+      # Typically there will be:
+      # - 1 "global" task that builds universal installers
+      # - N "local" tasks that build each platform's binaries and platform-specific installers
+      matrix: ${{ fromJson(needs.plan.outputs.val).ci.github.artifacts_matrix }}
+    runs-on: ${{ matrix.runner }}
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      BUILD_MANIFEST_NAME: target/distrib/${{ join(matrix.targets, '-') }}-dist-manifest.json
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - uses: swatinem/rust-cache@v2
+      - name: Install cargo-dist
+        run: ${{ matrix.install_dist }}
+      # Get the dist-manifest
+      - name: Fetch local artifacts
+        uses: actions/download-artifact@v4
+        with:
+          pattern: artifacts-*
+          path: target/distrib/
+          merge-multiple: true
+      - name: Install dependencies
+        run: |
+          ${{ matrix.packages_install }}
+      - name: Build artifacts
+        run: |
+          # Actually do builds and make zips and whatnot
+          cargo dist build ${{ needs.plan.outputs.tag-flag }} --print=linkage --output-format=json ${{ matrix.dist_args }} > dist-manifest.json
+          echo "cargo dist ran successfully"
+      - id: cargo-dist
+        name: Post-build
+        # We force bash here just because github makes it really hard to get values up
+        # to "real" actions without writing to env-vars, and writing to env-vars has
+        # inconsistent syntax between shell and powershell.
+        shell: bash
+        run: |
+          # Parse out what we just built and upload it to scratch storage
+          echo "paths<<EOF" >> "$GITHUB_OUTPUT"
+          jq --raw-output ".artifacts[]?.path | select( . != null )" dist-manifest.json >> "$GITHUB_OUTPUT"
+          echo "EOF" >> "$GITHUB_OUTPUT"
+
+          cp dist-manifest.json "$BUILD_MANIFEST_NAME"
+      - name: "Upload artifacts"
+        uses: actions/upload-artifact@v4
+        with:
+          name: artifacts-build-local-${{ join(matrix.targets, '_') }}
+          path: |
+            ${{ steps.cargo-dist.outputs.paths }}
+            ${{ env.BUILD_MANIFEST_NAME }}
+
+  # Build and package all the platform-agnostic(ish) things
+  build-global-artifacts:
+    needs:
+      - plan
+      - build-local-artifacts
+    runs-on: "ubuntu-20.04"
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      BUILD_MANIFEST_NAME: target/distrib/global-dist-manifest.json
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: Install cargo-dist
+        shell: bash
+        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/cargo-dist/releases/download/v0.11.1/cargo-dist-installer.sh | sh"
+      # Get all the local artifacts for the global tasks to use (for e.g. checksums)
+      - name: Fetch local artifacts
+        uses: actions/download-artifact@v4
+        with:
+          pattern: artifacts-*
+          path: target/distrib/
+          merge-multiple: true
+      - id: cargo-dist
+        shell: bash
+        run: |
+          cargo dist build ${{ needs.plan.outputs.tag-flag }} --output-format=json "--artifacts=global" > dist-manifest.json
+          echo "cargo dist ran successfully"
+
+          # Parse out what we just built and upload it to scratch storage
+          echo "paths<<EOF" >> "$GITHUB_OUTPUT"
+          jq --raw-output ".artifacts[]?.path | select( . != null )" dist-manifest.json >> "$GITHUB_OUTPUT"
+          echo "EOF" >> "$GITHUB_OUTPUT"
+
+          cp dist-manifest.json "$BUILD_MANIFEST_NAME"
+      - name: "Upload artifacts"
+        uses: actions/upload-artifact@v4
+        with:
+          name: artifacts-build-global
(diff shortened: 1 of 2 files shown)


ashleygwilliams:
to test this out we could push a pre-release tag and it would generate a github release. let me know if that sounds like a good plan and if you'd like to add or remove any targets or installers i picked.

Shnatsel:
Thanks for the PR! I'll take a closer look tomorrow.

Shnatsel:
Thanks! Two things of note:

1. We would need to also remove the build steps added in [#488](https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/488)
2. Is there an easy way to add aarch64 Linux as well? I'm pretty sure `cross` can handle that.

ashleygwilliams:
i can write up an additional workflow that could do aarch64 linux- cargo-dist doesn't currently support it (the workflow committed in this PR is fully generated, i can also hand edit the release workflow, but it's generally better to add an additional workflow that does the additional target). and happy to remove the other workflow.

Shnatsel:
In that case I'd prefer to stick to the fully generated workflow in the interest of reducing the maintenance burden and making upgrading to newer `cargo dist` easier. I don't want aarch64 linux prebuilt binaries badly enough to hand-roll them.

The previous release workflow would need to be removed, but other than that this looks good to go.

ashleygwilliams:
ok cool! also, now that we have a request for aarch64 linux we can look into getting it into cargo-dist :) i'll get the changes up momentarily for this PR

Shnatsel:
Thanks!

My thinking is that Raspberry Pis need prebuilt binaries really badly, and ARM servers are increasingly popular as well because of the way cloud providers price them. 

But it's not really needed for `cargo cyclonedx`. You would usually run `cargo cyclonedx` either on the developer machine or on CI, but RPis make terrible developer machines, and there is still no CI provider with easily available (let alone free) ARM servers.

lfrancke:
That said... We just build an ARM GitHub runner and would love this :) but yeah... Not too important

ashleygwilliams:
hey! realized i fell off on this and would love to pick it back up. will update to the latest cargo-dist now, and let me know what else i need to add to get this over the finish line! sorry about the delay

ashleygwilliams:
re linux arm- if there's a gustom runner we can point to i can add this as a target. at axo we use buildjet for this- costs us an incredibly small amount of money, would recommend (but also understand not wanting to).

Shnatsel:
The PR is still marked as draft. I understand it is ready for review now?

lfrancke:
@Shnatsel this is ready for review now. Would you be willing to take a look?

Shnatsel:
Yes! Thank you. I will take a look in the next few days.

stevespringett:
Any chance of getting a review of this PR?

Shnatsel:
Sorry for dragging my feet on the review.

Unfortunately this runs into some thorny issues with the way the crates are currently published. None of the active maintainers have access to the publishing token, and the original maintainers are incommunicado. This leaves us with publishing releases to crates.io using a token stored in a github action secret.

We recently had to add another crate to the project, because procedural macros require a separate crate for technical reasons. This complicates the release process even further. I want to sort out the release process with the extra added crate first, which is required for the crucial CycloneDX v1.5 support in the library, and only then deal with further less critical changes including this PR.

stevespringett:
@Shnatsel ping me on Slack and we'll sort it out.

Shnatsel:
There already is a [workflow](https://github.com/CycloneDX/cyclonedx-rust-cargo/blob/48139bd39681a2ced0098fcfa5d0d956b91151f5/.github/workflows/deploy_cargo_cyclonedx.yml#L52) we inherited from the previous maintainers that compiles and publishing binaries to Github releases. Having two of them at the same time is obviously not a good idea. Compiling binaries should be disentangled from the publishing workflow before I can land this PR.

I am being extra cautious with changes to the publishing workflow because without access to the publishing token we wouldn't be able to yank published crates, which makes correcting any mistakes in the release process very difficult.

ashleygwilliams:
a point that may be valuable to note: this current workflow does *not* include publishing to crates.io. i agree that extra consideration should be made there because it is a one way door. the pr included is for building binaries, and then also building a curl | sh installer, and a powershell installer, which i will need to do with you). these installers are preferable on CI platforms because they do not require compilation (which a cargo install does). we can also add a homebrew and npm installer if that's desirable with a bit of config, but that is not part of this PR.

everything in this PR generates something that is a two way door: all bins, installers, and github releases are mutable (deletable/fixable).

ashleygwilliams:
it is possible that a sync conversation about this may help any confusion. i am happy to make myself available on slack and/or a zoom call

Shnatsel:
I understand that this PR does not publish to crates.io. We have a [legacy workflow](https://github.com/CycloneDX/cyclonedx-rust-cargo/blob/main/.github/workflows/deploy_cargo_cyclonedx.yml) which does, and which also compiled binaries. The binary publishing part will have to be removed from the legacy workflow before this one can be merged. This should (hopefully) be as easy as reverting #488. Actually doing that is a prerequisite for landing this PR.

We already have a Github release with published binaries for the current version of `cargo cyclonedx`, so this PR will only take effect once we cut a new release. So here is my proposal for an actionable plan to get this merged:

1. Complete CycloneDX v1.5 support in `cyclonedx-bom` crate, tracked as #646 (@justahero)
2. Create a publishing workflow for `cyclonedx-bom-macros` crate and publish an initial release (@pvdrz, @Shnatsel)
3. Publish a new release of `cyclonedx-bom` crate (@Shnatsel)
4. Update `cargo cyclonedx` with spec version 1.5 support (@Shnatsel)
5. Revert #488 and double-check for any other edits required to the legacy publishing/tagging workflow (@Shnatsel)
6. Double-check the tag names in this PR and make sure everything aligns with the legacy publishing/tagging workflow, then merge this PR (@Shnatsel)
7. Publish a new version of `cargo cyclonedx`, finally making it accessible to `cargo binstall` et al.

My guesstimate for the ETA on all of this is 2 weeks.

I hope this sounds good?

Shnatsel:
We have sorted out the publishing permissions (thanks, Steve!) so the release process should be a lot less risky now.

I was about to ask how to go about upgrading to a newer version, but the cargo-dist book [already covers it](https://opensource.axo.dev/cargo-dist/book/updating.html). Nice!

ashleygwilliams:
@Shnatsel awesome news. let me know how i can help. additionally- if it comes up that there's targets or package manager systems this project needs that cargo-dist does not yet support, we are extremely open and excited about feature requests, so don't hesitate to let us know. 

once this is merged, my goal is to integrate cyclonedx BOM generation as a configurable feature. i've filed an issue to track work on it here: https://github.com/axodotdev/cargo-dist/issues/1016

Shnatsel:
Oh, SBOMs in cargo-dist sounds great! If you run into any issues with the tool, or have any questions, please let me know. I am happy to help!

ashleygwilliams:
also FYI- cargo-dist supports arm linux, the main thing youll need is a custom github runner (since we don't do it as a cross... yet). we use buildjet for cargo-dist and it costs us pennies (you can see our config here: https://github.com/axodotdev/cargo-dist/blob/main/Cargo.toml#L39). i dont know if owasp gives ya'll a budget, but i can set up the arm linux target if you'd like so long as you provide me a name for a custom runner!

ashleygwilliams:
anything more i can do to help on this? (hate to be a bother, but really hoping to use this for a demo)

Shnatsel:
I suppose I could trigger this workflow to build executables for an existing release without blocking on shipping CycloneDX 1.5.

The workflow is currently only triggered by pushing a version tag. I understand we would need to manually modify it to allow triggering it manually for a specific tag? 

If that is correct, I would be OK with merging a workflow with those modifications and triggering a binary release manually, but I'd like to revert to a more easily maintainable state without manual modifications to auto-generated files afterwards.

Shnatsel:
Alternatively I should be able to get the new release out of the door in a day or two.

Shnatsel:
I've opened a PR for shipping `cyclonedx-bom` v0.6.0: #710

Once that's done, I can finally revert #488, merge this PR, and trigger a build with the `cargo dist` workflow.

ashleygwilliams:
hey @Shnatsel there's a few options for changing how stuff is triggered without having to manually edit the generated workflow. if i'm understanding your concern correctly i think one of the following would work well, with the first being my preference:

- workflow dispatch ([docs](https://opensource.axo.dev/cargo-dist/book/reference/config.html#dispatch-releases)): this creates a tag at the end, and is only manually triggered. people who *REALLY* want to treat tags as immutable like this strategy. 
- tag namespace ([docs](https://opensource.axo.dev/cargo-dist/book/reference/config.html#tag-namespace)): if you want to temporarily require that the tags it triggers on have an arbitrary string as a namespace you could use this

ashleygwilliams:
also- would you like me to update the version of cargo-dist this uses? (we are also about to release 0.15.0 that has some cool features like more hash options)

Shnatsel:
Yes, an updated `cargo dist` would be great. Even just 0.14 would be nice to have.

Shnatsel:
The CI failure is due to an issue with the nightly compiler, already worked around in main. So I'm going to go ahead and merge this PR.

All right, let's give this a shot!

Shnatsel:
Well, the tag for v0.5.1 has been [created](https://github.com/CycloneDX/cyclonedx-rust-cargo/releases/tag/cargo-cyclonedx-0.5.1), but I don't think the release workflow has been triggered. Any idea why?

ashleygwilliams:
hrmmmm, let me see

ashleygwilliams:
ok yes! the reason is the tag structure - looking into how we can fix this, one sec

ashleygwilliams:
actually that's not correct at all. i am not sure what's going on, but i have to assume it's CI related- as everything works correctly locally, and we fully support the tag structure you have. continuing to investigate.

ashleygwilliams:
so i have been running some tests on our fork here: https://github.com/axodotdev/cyclonedx-rust-cargo/actions/runs/9197176872

i think that you are right- there is some sort of protection if the tag is originating from an action itself it seems?

ashleygwilliams:
ah @mistydemeo found this https://github.com/orgs/community/discussions/27028#discussioncomment-3254360 which seems to suggest that it's not possible to trigger one action from another

Shnatsel:
Okay, I see that I'll need to gut the release workflow and remove tagging from it.

In the meantime, how do I trigger the cargo-dist workflow for v0.5.1? Is deleting and recreating the tag sufficient?

ashleygwilliams:
apologies for the delay, yes deleting and recreating the tag will be sufficient to trigger the flow. so sorry for this oversight and the hassle that ensued! i have learned of some (arguably not pleasant) workarounds if you'd like to try to keep the tagging in the workflow. likely best to do manually for now but then we can discuss

Shnatsel:
I've manually recreated the tag, and the binary release is now up: https://github.com/CycloneDX/cyclonedx-rust-cargo/releases/tag/cargo-cyclonedx-0.5.1

However, I am seeing these warnings printed multiple times in the global artifact workflow:
`WARN !!! duplicate system keys, platforms may get conflated !!!`
see https://github.com/CycloneDX/cyclonedx-rust-cargo/actions/runs/9213590201/job/25348014566

I admit I do not understand what "system keys" are. Is this a misconfiguration on our side?

Also, the Markdown extracted from the changelog that goes onto the release page doesn't handle links correctly when the URL is provided [separately](https://github.com/CycloneDX/cyclonedx-rust-cargo/blob/8ababe9343e7387eab35ed32f5e0dd5f96685f7f/cargo-cyclonedx/CHANGELOG.md?plain=1#L97) instead of inline. Note the [release page](https://github.com/CycloneDX/cyclonedx-rust-cargo/releases/tag/cargo-cyclonedx-0.5.1) showing `([#559])` while the [full rendered changelog](https://github.com/CycloneDX/cyclonedx-rust-cargo/blob/main/cargo-cyclonedx/CHANGELOG.md) shows `(#559)`. This is not a big deal for us since we're just linking to PRs anyway, but it figured I'd let you know.

ashleygwilliams:
thanks for letting me know about the markdown thing. we are not formally parsing the markdown so i wonder if this is a github thing. regardless i'll take a further look. i know that we have users who do similar style linking, but we haven't seen this so i'm curious what's going on.

as for that warning... i have never seen that in the wild. only development, and very early on. i'm running a test right now (https://github.com/axodotdev/cyclonedx-rust-cargo/actions/runs/9213869039). i am wondering if it is related to the bumpy tagging situation. once i can repro i'll dig in further. your release appears completely fine and the installers work great so i'm hoping it's a false positive. here's where in our code it is triggered: https://github.com/axodotdev/cargo-dist/blob/632d7584d0732c2ee10e19561f168811c4f92c38/cargo-dist/src/manifest.rs#L132

ashleygwilliams:
ok, confirmed. that warning is a false positive, you can safely ignore and we'll remove it for the next release

Shnatsel:
Thanks for looking into the warning!

> i know that we have users who do similar style linking, but we haven't seen this so i'm curious what's going on.

I think your code just grabs everything under the appropriate header. Our changelog has the URLs for the links at the end of the file, so the URLs do not get included into the snippet posted to the release page.

But this is an edge case, and I'm not sure it's worth the trouble of introducing a markdown parser just to get such style of links to work. Especially since Markdown is so loosely specified and all the parsers end up being subtly incompatible anyway.

Shnatsel:
Now that the release binaries are published, it should be easy to integrate `cargo cyclonedx` into `cargo dist`. Please let me know if you have any questions or encounter any issues. I am happy to help!

Speaking of SBOMs, there are people requesting `cargo auditable` support in `cargo dist` too: https://github.com/axodotdev/cargo-dist/issues/81
I assume having published binaries would simplify integration into `cargo dist`, so I'm going to go ahead and set that up as well.

ashleygwilliams:
oh that's super awesome. yes cargo auditable integration would definitely be welcome! will let you know. i dont know if you use discord but we have one: https://discord.gg/hucxPFB6. otherwise i'm honestly happy to just keep using this PR as a convo space :)

Shnatsel:
@ashleygwilliams I've published a new release, v0.5.2, and `cargo binstall` no longer picks up on binaries from our GH releases made with `cargo dist`:

```
$ cargo binstall cargo-cyclonedx
 WARN Failed to retrieve token from `gh auth token` err=Os { code: 2, kind: NotFound, message: "No such file or directory" }
 WARN Failed to read git credential file
 INFO resolve: Resolving package: 'cargo-cyclonedx'
 WARN The package cargo-cyclonedx v0.5.2 will be installed from source (with cargo)
Do you wish to continue? yes/[no]
? 
```

If I use `--version=0.5.1` it downloads a binary from a third-party build service, so we get a binary but it isn't sourced from our Github releases:

```
cargo binstall --version=0.5.1 cargo-cyclonedx
 WARN Failed to retrieve token from `gh auth token` err=Os { code: 2, kind: NotFound, message: "No such file or directory" }
 WARN Failed to read git credential file
 INFO resolve: Resolving package: 'cargo-cyclonedx'
 INFO resolve: Verified signature for package 'cargo-cyclonedx-0.5.1-x86_64-unknown-linux-gnu': timestamp:1716442392	file:cargo-cyclonedx-0.5.1-x86_64-unknown-linux-gnu.tar.gz	hashed
 WARN The package cargo-cyclonedx v0.5.1 (x86_64-unknown-linux-gnu) has been downloaded from third-party source QuickInstall
 INFO This will install the following binaries:
 INFO   - cargo-cyclonedx (cargo-cyclonedx -> /home/shnatsel/.cargo/bin/cargo-cyclonedx)
Do you wish to continue? yes/[no]
? 
```

Which leads me to believe that `cargo dist` never really worked with `cargo binstall` on our repo after all.

Could you take a look and see what's up with that?

ashleygwilliams:
@Shnatsel oh huh! that is very weird, and yeah we'll take a look right now. cargo binstall works on cargo-dist and so i wonder if there's something perhaps with tag structure or the workspace that is making the lookup fail

ashleygwilliams:
ok! so the issue is that your tag structure is not one of the two default tag structures that binstall looks up. if you run `cargo binstall cargo-cyclonedx --dry-run --verbose` you can see this:

```
DEBUG Checking for package at: 'https://github.com/CycloneDX/cyclonedx-rust-cargo/releases/download/0.5.2/cargo-cyclonedx-x86_64-pc-windows-msvc.zip'

DEBUG Checking for package at: 'https://github.com/CycloneDX/cyclonedx-rust-cargo/releases/download/v0.5.2/cargo-cyclonedx-x86_64-pc-windows-msvc.zip'
```

luckily, binstall lets you write config ([docs](https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md)) to change the structure of the tag. you'll probably want something like this:

```
pkg-url = "{ repo }/releases/download/cargo-cyclonedx-{ version }/{ name }-{ target }{ archive-suffix }"
```

you can make sure this is 100% correct by running the dry run command from above.

because binstall uses crates.io metadata as their source of truth, this config won't apply retroactively to previous version but all future versions with this config should work.

ashleygwilliams:
(sorry for not catching this sooner, we should honestly doc this better on our end and will do so!)

Shnatsel:
Yeah, it would be neat if `cargo dist` would print a warning when the tag structure doesn't match what `cargo binstall` expects, and point me to the cargo-binstall documentation.

The config snippet you provided does indeed work. I've opened a PR to add it: https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/727

Thanks a lot!

ashleygwilliams:
@Shnatsel yeah that's an interesting idea. my thought was honestly wondering if cargo dist init could write the binstall config for you. this would require knowing your tag structure, so i'll have to think through how we might want to do that, but i agree there's some work here to make this a nicer experience!!

Shnatsel:
It could also verify that it matches in CI on pull requests. There's already tag verification at that stage.

PR #558, opened 2023-11-15
add --version flag

tshepang:
Closes #555

PR #561, opened 2023-11-16
Add `--version` support

Shnatsel:
#558 but with all the requested changes

Fixes https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/555


tshepang (approved):

tshepang:
On a related note (perhaps in a future PR), we should move to non-deprecated names, where we replace `clap` with `command` and `arg`. Building with `deprecated` feature of clap will give guidance where this should happen.

Issue #532, opened 2023-11-02
Capture data only available during the build process


Shnatsel:
There is a certain amount of data that is only available during or after the actual build, such as a hash of the resulting binary, the `RUSTFLAGS` used (not just from the environment variable but also from the Cargo configuration), etc.

There are two ways to implement it:

1. Hook into the build process, similar to how https://github.com/rust-secure-code/cargo-auditable works
2. The hypothetical "build info" file that's being discussed by the Cargo team could be used to achieve this.

Shnatsel:
`cargo build -Z unstable-options --build-plan ` _might_ also help. I have not investigated it in detail.

lfrancke:
There is a [ticket](https://github.com/rust-lang/cargo/issues/7614) to maybe delete build-plan support so I would not recommend relying on it.

In any case it is the plan on what will run not what did run.
The idea of @edpage @arlosi was to introduce a machine readable file ("build info") that records everything that _actually_ was built (and how).

There is no tracking issue for this yet but I hope that we'll have something soon.

Shnatsel:
The build-info pre-RFC can be found here: https://internals.rust-lang.org/t/pre-rfc-cargo-sbom/19842

PR #563, opened 2023-11-16
Add an option to write a SBOM per binary

### LINKED ISSUE #557: Provide an easy way to map an artifact to a SBOM
A rust build can produce various artifacts in the `/target` folders.
The SBOMs might have different names (e.g. `bom.xml` or the name of the package) and it's not trivial (without parsing the SBOMs) to automatically map/match the artifact to the SBOM in question.

A second, related issue (I can split it into a separate issue) is that the `metadata -> component -> name` might be an "internal artifact" and not the name of the actual final product as used in the outside world (e.g. vulnerabilities might be filed against the "official" name or the "binary" name but not the package/crate name).

The "correct" name might be available as a subcomponent of `metadata -> component` but it's not obvious to tools which the actual thing is we're talking about (e.g. a vulnerabilty might exist in a binary but not in the library).

The only thing I can think of is to actually do create a separate SBOM file per generated artifact. That is overhead and duplication but I don't see a better way of doing this.

This was my attempt at doing this in the past and it might still be what I need: https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/441

Shnatsel:
Mostly (?) fixes #557

This sticks to having the whole SBOM operate in terms of packages, so that the toplevel element is still a package potentially consisting of several binaries; but there is a SBOM named after every binary, making them much easier to locate.

### Why do it this way

So you may be thinking "Eww, gross! Why don't you just make the binary your toplevel element? Why generate a SBOM for the whole package?"

There is a long list of reasons why it is done this way, most of which are weaksauce ("all components are packages with PURLs so it's more consistent", "the per-binary mode isn't special and isn't a source of lots of extra complexity and bugs", etc). The **real** reason we output a SBOM for the whole package is **the SBOMs clobbering each other.**

Imagine you have a target that's both a binary and a shared library. The compiled artifacts get different extensions (`.exe` vs `.dll` on Windows, nothing vs `.so` on Linux), etc. However, when you try to emit stuff like debug info for them, the names will clash! Both are `myproject.pdb` on Windows. If you run into this, Cargo will print a warning but not actually do anything to fix this.

We have to deal with this problem somehow as well.

There are two sources of clobbering:

1. The target with the same name (this is common)
2. The workspace producing e.g. a library and a binary with the same name but coming from different packages (also common), or even binaries with the same name (unlikely)

We easily sidestep (2) by writing the SBOM into each package's directory. Boom, stuff from different packages cannot clash.

We sidestep (1) by emitting the SBOM for the whole package. So the library and binary targets within the same package can clobber each other as much as they like - it's still going to be the exact same file! You don't have to invent a scheme to differentiate between an rlib, a cdylib and an executable that all got built from the same package - you just look up the name of the binary and off you go.

In case you're wondering why don't we just name each SBOM after the final name of the binary, `.exe` and all: the answer is that we don't know them. `cargo metadata` does not provide this information. https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/532 might help with that. But even if that ever happens, that would still be incompatible with the `--target=all` mode.

------

You may notice that these things are now named after a variable prefix and no longer fit the standard naming convention. And that is absolutely correct! That's a bug that was present all along for `--output-pattern=package` and that we inherited in this PR too. I think there needs to be a `.cdx` in there when it's not `bom.{xml,json}`. That is a change in user-visible behavior so that'll require shipping a 0.5.0 release. I think this is a necessary fix, I'll make a PR for that once this is merged to avoid conflicts.

### DIFF
--- a/cargo-cyclonedx/src/cli.rs
+++ b/cargo-cyclonedx/src/cli.rs
@@ -79,7 +79,7 @@ Defaults to the host target, as printed by 'rustc -vV'"
     #[clap(long = "output-cdx")]
     pub output_cdx: bool,
 
-    /// Prefix patterns to use for the filename: bom, package
+    /// Prefix patterns to use for the filename: bom, package, binary
     #[clap(
         name = "output-pattern",
         long = "output-pattern",
--- a/cargo-cyclonedx/src/config.rs
+++ b/cargo-cyclonedx/src/config.rs
@@ -164,6 +164,7 @@ pub enum Pattern {
     #[default]
     Bom,
     Package,
+    Binary,
 }
 
 impl FromStr for Pattern {
--- a/cargo-cyclonedx/src/config.rs
+++ b/cargo-cyclonedx/src/config.rs
@@ -175,7 +175,10 @@ impl FromStr for Pattern {
             "bom" => Ok(Self::Bom),
             "package" => Ok(Self::Package),
             "binary" => Ok(Self::Binary),
-            _ => Err(format!("Expected bom or package, got `{}`", s)),
+            _ => Err(format!(
+                "Expected 'bom', 'package' or 'binary', got `{}`",
+                s
+            )),
         }
     }
 }
--- a/cargo-cyclonedx/src/config.rs
+++ b/cargo-cyclonedx/src/config.rs
@@ -174,6 +174,7 @@ impl FromStr for Pattern {
         match s {
             "bom" => Ok(Self::Bom),
             "package" => Ok(Self::Package),
+            "binary" => Ok(Self::Binary),
             _ => Err(format!("Expected bom or package, got `{}`", s)),
         }
     }
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -631,6 +631,7 @@ impl GeneratedSbom {
         let prefix = match output_options.prefix {
             Prefix::Pattern(Pattern::Bom) => "bom".to_string(),
             Prefix::Pattern(Pattern::Package) => self.package_name.clone(),
+            Prefix::Pattern(Pattern::Binary) => todo!(),
             Prefix::Custom(c) => c.to_string(),
         };
 
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -53,9 +53,10 @@ use regex::Regex;
 
 use log::Level;
 use std::collections::BTreeMap;
+use std::collections::BTreeSet;
 use std::convert::TryFrom;
 use std::fs::File;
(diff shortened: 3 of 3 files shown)


cargo-cyclonedx/src/config.rs
```diff
@@ -173,6 +174,7 @@ impl FromStr for Pattern {
         match s {
             "bom" => Ok(Self::Bom),
             "package" => Ok(Self::Package),
+            "binary" => Ok(Self::Binary),
             _ => Err(format!("Expected bom or package, got `{}`", s)),
```
justahero, line 178: I suggest to update the error message to something like:

```rust
_ => Err(format!("Expected 'bom', 'package' or 'binary', got `{}`", s))
```

justahero (approved):
Changes look good to me.

cargo-cyclonedx/src/cli.rs
```diff
@@ -80,7 +80,7 @@ Defaults to the host target, as printed by 'rustc -vV'"
     #[clap(long = "output-cdx")]
     pub output_cdx: bool,
 
-    /// Prefix patterns to use for the filename: bom, package
+    /// Prefix patterns to use for the filename: bom, package, binary
```
lfrancke, line 83: Can we not add 
"Implies --output-cdx" for binary?

lfrancke:
I'm afraid this returns the wrong result. At least not what I'd expect and in this form it'd not be too helpful.

If I have a member with multiple targets (e.g. multiple binaries) I'd expect the top-level  `metadata.component` to be exactly that.
e.g. I have a crate called cargo-cyclonedx with a `main.rs` plus `bin/hello_world.rs` and `bin/hello_world2.rs` I'd expect three SBOMs.
One called `hello_world.cdx.json` which would have a top-level `metadata.component` with the name `hello_world` and not `cargo-cylonedx`.

We can ask this on the CycloneDX slack but that'd be my interpretation of what users will expect.
Libraries don't get any separate SBOMs in this case either.

The implementation might suck but my [original PR](https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/441) implemented what I thought was sane back then.

Shnatsel:
I see that this does not fit your use case. I'm glad I held off merging until I got your feedback. I'm going to close this, and work on solving your use case in another PR.

----

>  e.g. I have a crate called cargo-cyclonedx with a main.rs plus bin/hello_world.rs and bin/hello_world2.rs I'd expect three SBOMs.

To clarify, this is the case in this PR.

>  One called hello_world.cdx.json which would have a top-level metadata.component with the name hello_world and not cargo-cylonedx.

But this is not.

>  Libraries don't get any separate SBOMs in this case either.

This is not the case in this PR either.

PR #577, opened 2023-12-05
Revert Licenses & LicenseChoice changes

justahero:
This reverts the changes of types `Licenses` & `LicenseChoice` to a previous state. In a previous commit they changed to:

```rust
pub(crate) struct Licenses(pub LicenseChoice);

pub(crate) enum LicenseChoice {
    Licenses(Vec<Lic>),
    Expression(Vec<Expr>),
}
```

This commit reverts the change back to

```rust
pub(crate) struct Licenses(pub Vec<LicenseChoice>);

pub(crate) struct LicenseChoice {
    License(License),
    Expression(String),
}
```

**Update** the models have been reverted to accomodate older CycloneDX documents.


Shnatsel:
Okay, I understand now where the confusion arose.

In 1.4 [multiple expressions are permitted](https://cyclonedx.org/docs/1.4/json/#tab-pane_components_items_licenses_items_oneOf_i0), which is proposed in this PR.

But in 1.5 it's [either a single SPDX expression or a list of licenses](https://cyclonedx.org/docs/1.5/json/#components_items_licenses), to which `main` has already been converted.

cyclonedx-bom/src/models/license.rs
```diff
@@ -208,24 +237,52 @@ mod test {
...
-        )]))
+        ))])
         .validate_with_context(ValidationContext::default())
         .expect("Error while validating");
 
         assert_eq!(validation_result, ValidationResult::Passed);
     }
 
+    #[test]
+    fn it_should_fail_with_mixed_license_nodes() {
```
Shnatsel, line 250: This test (and the associated code) should only be enabled for 1.5, not 1.4.

Shnatsel (changes requested):
1.4 allows mixing the licenses and expressions, while 1.5 does not.

We need to support loading CycloneDX documents created when this was allowed, so we need to use the old data model (as proposed by this PR) and inject the check on loading and saving 1.5 only.

PR #588, opened 2023-12-15
Add support for `signature` field in 1.4

justahero:
This PR adds support for the [signature](https://cyclonedx.org/docs/1.4/json/#signature) field. The signature field can appear in three different versions, for now only the flat [`Signature` option](https://cyclonedx.org/docs/1.4/json/#tab-pane_signature_oneOf_i2) is implemented with required fields `algorithm` & `value` to get started. Support for the other versions can follow later, the internal Signature struct would need to change to support multiple options (by using an enum) but unfortunately have not found a good way to do that at this point.

The `signature` field needs to be added to the following nodes in the spec: `components`, `services` & `compositions`. The validation of the provided algorithm needs to move out from the parse logic.

* add `Signature` & `Algorithm` types with serialization / deserialization logic for JSON and XML
* add `signature` field to `Bom` & `Component` (metadata) structs, is ignored in version 1.3 (set to `None`)
* adjust samples / snapshots to include new signature field
* update tests
* add comments to signal which fields are new in version 1.4

### DIFF
--- a/cyclonedx-bom/src/models/signature.rs
+++ b/cyclonedx-bom/src/models/signature.rs
@@ -1,0 +1,115 @@
+/*
+ * This file is part of CycloneDX Rust Cargo.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use std::str::FromStr;
+
+/// Enveloped signature in [JSON Signature Format (JSF)](https://cyberphone.github.io/doc/security/jsf.html)
+#[derive(Debug, PartialEq, Eq)]
+pub struct Signature {
+    /// Signature algorithm.
+    pub algorithm: Algorithm,
+    /// The signature data.
+    pub value: String,
+}
+
+/*
+/// Enveloped signature in [JSON Signature Format (JSF)](https://cyberphone.github.io/doc/security/jsf.html)
+#[derive(Debug, PartialEq, Eq)]
+pub enum Signature {
+    /// Multiple signatures
+    Signers(Vec<Signer>),
+    /// A single signature chain
+    Chain(Signer),
+    /// A single signature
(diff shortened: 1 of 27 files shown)


cyclonedx-bom/src/models/signature.rs
```diff
@@ -0,0 +1,115 @@
...
+
+/// For now the [`Signer`] struct only holds algorithm and value
+#[derive(Debug, PartialEq, Eq)]
+pub struct Signer {
+    /// Signature algorithm.
+    pub algorithm: Algorithm,
+    /// The signature data.
+    pub value: String,
+}
+*/
```
Shnatsel, line 50: What's up with the commented-out code? Ideally it should be either removed, or have a really good explanation why it's disabled in a comment next to it.

cyclonedx-bom/src/models/signature.rs
```diff
@@ -0,0 +1,115 @@
...
+            "PS512" => Ok(Algorithm::PS512),
+            "ES256" => Ok(Algorithm::ES256),
+            "ES384" => Ok(Algorithm::ES384),
+            "ES512" => Ok(Algorithm::ES512),
+            "Ed25519" => Ok(Algorithm::Ed25519),
+            "Ed448" => Ok(Algorithm::Ed448),
+            "HS256" => Ok(Algorithm::HS256),
+            "HS384" => Ok(Algorithm::HS384),
+            "HS512" => Ok(Algorithm::HS512),
+            _ => Err(format!("Invalid algorithm '{}' found", s)),
```
Shnatsel, line 90:
```suggestion
            _ => Err(format!("Invalid signature algorithm '{}' found", s)),
```

cyclonedx-bom/src/specs/v1_4/signature.rs
```diff
@@ -0,0 +1,336 @@
...
+            "PS512" => Ok(Algorithm::PS512),
+            "ES256" => Ok(Algorithm::ES256),
+            "ES384" => Ok(Algorithm::ES384),
+            "ES512" => Ok(Algorithm::ES512),
+            "Ed25519" => Ok(Algorithm::Ed25519),
+            "Ed448" => Ok(Algorithm::Ed448),
+            "HS256" => Ok(Algorithm::HS256),
+            "HS384" => Ok(Algorithm::HS384),
+            "HS512" => Ok(Algorithm::HS512),
+            _ => Err(format!("Invalid algorithm '{}' found", s)),
```
Shnatsel, line 96: ditto

Shnatsel (commented):
Looks good by and large, just needs some minor tweaks and then I can merge it.

Note to self: this requires a semver breaking change due to addition of new `pub` fields on `pub struct`s that aren't `#[non_exhaustive]`

Shnatsel:
I'm going to merge this to unblock my own work and avoid creating conflicts, but I expect the issue of the commented out code to be resolved.

PR #606, opened 2024-01-31
Refactor initialization of regexes

justahero:
Instead of raising an `RegexError` when regular expressions are initialized they are now expected to be compiled successfully, otherwise the code will panic. All these regexes are covered by tests, if one of them fails to compile an associated test needs to fail.

One reason for this change is that these regular expressions should be treated as a compile time concern, rather than checking them during runtime. Otherwise we might release a version that contains an invalid & untested regex, that short cuts the validation logic without any chance to remedy it. If a regex fails to compile it should fail immediately in a test.

The second reason for this change is to prepare a refactor of the validation logic. Currently the only way a validation method can return a `ValidationError` is when one of the regexes fails to compile during runtime. Regexes are expected to be valid in their functions, the `Validate` trait signals an improperly used return type. Validation errors are all collected inside the `ValidationResult`, not in the `ValidationError` enum.

### DIFF
--- a/cargo-cyclonedx/tests/cli.rs
+++ b/cargo-cyclonedx/tests/cli.rs
@@ -75,6 +75,7 @@ fn find_content_in_bom_files() -> Result<(), Box<dyn std::error::Error>> {
 }
 
 #[test]
+#[ignore]
 fn find_content_in_stderr() -> Result<(), Box<dyn std::error::Error>> {
     let tmp_dir = make_temp_rust_project()?;
 
--- a/cyclonedx-bom/src/models/bom.rs
+++ b/cyclonedx-bom/src/models/bom.rs
@@ -516,8 +516,8 @@ pub struct UrnUuid(pub(crate) String);
 impl UrnUuid {
     pub fn new(value: String) -> Result<Self, UrnUuidError> {
         match matches_urn_uuid_regex(&value) {
-            Ok(true) => Ok(Self(value)),
-            Ok(false) | Err(_) => Err(UrnUuidError::InvalidUrnUuid(
+            true => Ok(Self(value)),
+            false => Err(UrnUuidError::InvalidUrnUuid(
                 "UrnUuid does not match regular expression".to_string(),
             )),
         }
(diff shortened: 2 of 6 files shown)


cargo-cyclonedx/tests/cli.rs
```diff
@@ -75,6 +75,7 @@ fn find_content_in_bom_files() -> Result<(), Box<dyn std::error::Error>> {
 }
 
 #[test]
+#[ignore]
```
pvdrz, line 78: why does this test needs to be ignored?

cyclonedx-bom/src/models/bom.rs
```diff
@@ -563,15 +562,12 @@ pub enum UrnUuidError {
     InvalidUrnUuid(String),
 }
 
-fn matches_urn_uuid_regex(value: &str) -> Result<bool, regex::Error> {
-    static UUID_REGEX: Lazy<Result<Regex, regex::Error>> = Lazy::new(|| {
+fn matches_urn_uuid_regex(value: &str) -> bool {
```
pvdrz, line 565: given that you had to change this code, it could be nice if `matches_urn_uuid_regex` returns a `ValidationResult` directly so `UrnUuid::new` and `UrnUuid::validate_with_context` don't have to map the `bool` to a `ValidationResult` themselves.

I wouldn't do this if `matches_urn_uuid_regex` is used elsewhere without mapping the `bool`.

cyclonedx-bom/src/models/component.rs
```diff
@@ -460,26 +455,21 @@ impl Validate for Cpe {
...
-                    Ok(ValidationResult::Failed {
-                        reasons: vec![FailureReason {
-                            message: "Cpe does not match regular expression".to_string(),
-                            context,
-                        }],
-                    })
-                }
-            }
-            Err(e) => Err(e.clone().into()),
+        if UUID_REGEX.is_match(&self.0) {
```
pvdrz, line 464: maybe this could be a `match { true => /* ... */, false => /* ... */}` instead to be consistent with the rest of the code, wdyt?

pvdrz (approved):
This looks great to me, I left some minor comments but they aren't blocking.

I'd like to also mention that the [`clippy::invalid_regex`](https://rust-lang.github.io/rust-clippy/master/#/regex) lint should already catch any invalid regex on "compile-time". This lint is in the `deny` category by default so we don't have to do anything so it catches any invalid regex.

pvdrz:
I also wonder if `ValidationResult` should have a method like
```rust
impl ValidationResult {
    fn new(v: bool, or_else: impl FnOnce() -> ValidationResult) {
        match v {
            true -> ValidationResult::Passed,
            false -> or_else(),
        }
    }
}
```
To deduplicate all those `match` everywhere. But I'm not entirely convinced. I'd do that in a different PR though.

cyclonedx-bom/src/models/bom.rs
justahero: I think it's ok to keep it as is, function `matches_urn_uuid_regex` only checks if the given string matches, while the validation method maps that to the ValidationResult. The alternative could have been to merge both into one function.

cargo-cyclonedx/tests/cli.rs
justahero: This is my mistake, this is the only test that fails on my machine (I have not checked yet why).

PR #609, opened 2024-02-01
Refactor `Validate` trait to simplify return type

justahero:
This PR changes the methods of the `Validate` trait to not return a `Result`.

Before this change the `Validate` trait looked as follows:

```rust
pub trait Validate {
    fn validate_with_context(
        &self,
        context: ValidationContext,
    ) -> Result<ValidationResult, ValidationError>;
}
```

Due to PR #606 all occurrences of regular expressions have been refactored to not return `RegexError` anymore. A `ValidationError` was only returned when a regular expression failed to compile during runtime. Tests now ensure that an invalid Regex is caught immediately & all regexes are valid. This eliminates the requirement for `validate_with_context` to return a `Result`. All validation errors are aggregated in the `ValidationResult` type.

The trait was simplified to:

```rust
pub trait Validate {
    fn validate_with_context(&self, context: ValidationContext) -> ValidationResult;
}
```

Unfortunately this change affects a lot of validation methods for a lot of different structs.

* refactor `validate` & `validate_with_context` implementations
* refactor tests to use `validate()` rather than `validate_with_context(ValidationContext::default())`

### DIFF
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -99,7 +99,7 @@ impl SbomGenerator {
             let bom = generator.create_bom(member, &dependencies, &pruned_resolve)?;
 
             if cfg!(debug_assertions) {
-                let result = bom.validate().unwrap();
+                let result = bom.validate();
                 if let ValidationResult::Failed { reasons } = result {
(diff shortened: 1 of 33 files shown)


Shnatsel:
Looks good to me, but CI is failing.

Issue #615, opened 2024-02-07
CycloneDx 1.4 requires "dependsOn" but it is not mandatory in the spec.


carlosthe19916:
- The `dependencies.dependsOn` field is not mandatory but optional. According to the spec defined here https://cyclonedx.org/docs/1.4/json/#dependencies_items_dependsOn
- While parsing an SBOM that contains `dependencies` without `dependsOn` I got an error `JsonElementReadError { error: Error("missing field `dependsOn`", line: 0, column: 0) }`

### Reproduce the issue 
- Clone this small demo repository https://github.com/carlosthe19916/cyclonedx14
- Execute `cargo run`. We will ran into an error.

### Some context
- I have this demo SBOM (CycloneDx) v.1.4 https://github.com/carlosthe19916/cyclonedx14/blob/master/resources/bom.1.4.json
- I want to read that file using
```
cyclonedx_bom::prelude::Bom::parse_from_json(file)
```
- An error is generated
```
JsonElementReadError { error: Error("missing field `dependsOn`", line: 0, column: 0) }
```

The SBOM contains `dependencies` without `dependsOn` field, which is acceptable by the spec but not accepted by the library

Is this a bug or the expected behavior of the library?
If more information is needed I'm happy to help.

PR #616, opened 2024-02-08
fix: Make CycloneDx dependencies.dependsOn optional

carlosthe19916:
Resolves https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/615

Hi, I reported the issue above. And I though it would be good to help with it. Although I'm not sure of whether or not this is a good solution I'd be happy if you could give it a review and let me know if that is the way to solve this issue.

### DIFF
--- a/cyclonedx-bom/src/specs/v1_3/dependency.rs
+++ b/cyclonedx-bom/src/specs/v1_3/dependency.rs
@@ -83,6 +83,7 @@ impl FromXml for Dependencies {
 pub(crate) struct Dependency {
     #[serde(rename = "ref")]
     dependency_ref: String,
+    #[serde(default)]
     depends_on: Vec<String>,
 }
 
--- a/cyclonedx-bom/src/specs/v1_4/dependency.rs
+++ b/cyclonedx-bom/src/specs/v1_4/dependency.rs
@@ -83,6 +83,7 @@ impl FromXml for Dependencies {
 pub(crate) struct Dependency {
     #[serde(rename = "ref")]
     dependency_ref: String,
+    #[serde(default)]
     depends_on: Vec<String>,
 }
 


Shnatsel:
Thank you!

@justahero could you take a look? I'd appreciate your review since you're more involved with the `cyclonedx-bom` side of things.

justahero:
Hey @carlosthe19916 , thanks for pointing this out. :+1:

cyclonedx-bom/src/specs/v1_3/dependency.rs
```diff
@@ -83,14 +83,14 @@ impl FromXml for Dependencies {
 pub(crate) struct Dependency {
     #[serde(rename = "ref")]
     dependency_ref: String,
-    depends_on: Vec<String>,
+    depends_on: Option<Vec<String>>,
```
justahero, line 86: If you annotate the `depends_on` field with [serde's `default` attribute](https://serde.rs/attr-default.html)
```suggestion
    #[serde(default)]
    depends_on: Vec<String>,
```

an empty `Vec` (default) is set instead in case the field is absent from the JSON input. Then all other changes can be reverted.

justahero: This is for version 1.3. Would it be ok to apply the change to the 1.4 version of [Dependency](https://github.com/CycloneDX/cyclonedx-rust-cargo/blob/main/cyclonedx-bom/src/specs/v1_4/dependency.rs#L86) too as the behavior according to the [Spec 1.4](https://cyclonedx.org/docs/1.4/json/#dependencies_items_dependsOn) is the same?

carlosthe19916: Thanks for your review @justahero ! I've applied your suggested change.

justahero (approved):

PR #618, opened 2024-02-12
Update 'version' to be required field

justahero:
According to the JSON specification (in [1.3](https://cyclonedx.org/docs/1.3/json/#version) & [1.4](https://cyclonedx.org/docs/1.4/json/#version)) the top-level field 'version' is required, it defaults to 1.

Therefore the field is refactored to not be optional.

* update conversion & serialization functions

### DIFF
--- a/cyclonedx-bom/src/specs/v1_3/bom.rs
+++ b/cyclonedx-bom/src/specs/v1_3/bom.rs
@@ -44,7 +44,7 @@ use xml::{reader, writer::XmlEvent};
 pub(crate) struct Bom {
     bom_format: BomFormat,
     spec_version: SpecVersion,
-    version: Option<u32>,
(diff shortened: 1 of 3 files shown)


Shnatsel:
Now that I am actually using this in `cargo auditable`, it seems that this is not a good idea. Making it required results in it always being written to the BOM, which prevents clean roundtrips. And `version: 1` is redundant because the spec states it should be treated as 1 if not present, and takes up valuable space in the `cargo auditable` use case.

I would like to revert this in the next semver-breaking release, to give the caller more control over the serialized SBOM and enable cleaner roundtrips with fewer alterations. I would also like to provide an accessor function instead that returns the version if set or 1 if not in the next release, to get easy access to the spec-compliant logic.

PR #619, opened 2024-02-17
BOM per binary, take 3

Shnatsel:
Adds `--output-pattern=binary` and `--output-pattern=cargo-target` modes that emit SBOMs for compiled binaries and for all compilation targets (including Rust libraries that do not exist as standalone binaries) respectively.

@lfrancke please test this and let me know if `--output-pattern=binary` fulfills your needs.

### DIFF
--- a/cargo-cyclonedx/src/cli.rs
+++ b/cargo-cyclonedx/src/cli.rs
@@ -80,7 +80,7 @@ Defaults to the host target, as printed by 'rustc -vV'"
     #[clap(long = "output-cdx")]
     pub output_cdx: bool,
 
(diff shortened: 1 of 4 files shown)


lfrancke (approved):
Looking good to me.
I tried it on two repositories and the results look good.

When I had my PR for his (naive in comparison) I used the "lib" as a prefix so that the name of the artifact would look similar to the name of the SBOM.
But I also found out that the final name of an artifact is not currently exposed anywhere.

This might come with https://github.com/rust-lang/rfcs/pull/3553
Anyway...good to go as is I think. There is no "correct" way to name these files so either way is good.

Just needs your DCO

PR #620, opened 2024-02-18
Record package hashes in the generated SBOM

Shnatsel:
Requires Rust 1.77 or later because it relies on https://github.com/rust-lang/cargo/pull/12914

Older Rust versions behave as before, i.e. the hashes are not recorded.

Builds on  #619 because I didn't want to create editing conflicts with myself.  I can rebase it if it doesn't go in.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -146,6 +146,7 @@ dependencies = [
  "anyhow",
  "assert_cmd",
  "assert_fs",
+ "cargo-lock",
  "cargo_metadata",
  "clap",
  "cyclonedx-bom",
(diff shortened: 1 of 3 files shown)


lfrancke (approved):
I looked at the code changes and those look good.
I only have 1.76 installed and train wifi won't allow me to download a newer version.
I'm happy to merge as is but if you prefer I can also test it next week.

Shnatsel:
In my testing it works, so I'm going to go ahead and merge it.

There is an edge case: if the registry URL contains parameters, the hash may not be recorded. That's because [Cargo.lock does not encode parameters correctly right now](https://github.com/rust-lang/cargo/pull/12852) (and what `cargo metadata` and `cargo pkgid` do is anyone's guess), but we'll get to that once the v4 `Cargo.lock` format which fixes it gets actually rolled out.

Shnatsel:
I filed an issue about that so that it's tracked: #629

PR #623, opened 2024-02-21
cargo update a day keeps dependabot away

Shnatsel:


Shnatsel:
Ah, Clap and MSRV again :disappointed:

PR #630, opened 2024-02-27
More robust target kind recording

Shnatsel:
Refactor target kind logic to remove dependency on the iteration order, making it less fragile.

### DIFF
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -76,6 +76,15 @@ pub struct SbomGenerator {
     crate_hashes: HashMap<cargo_metadata::PackageId, Checksum>,
 }
 
+/// Contains a map from `bom_ref` of a subcomponent to the kinds of Cargo targets it has,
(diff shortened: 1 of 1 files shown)


cargo-cyclonedx/src/generator.rs
```diff
@@ -757,7 +766,8 @@ impl GeneratedSbom {
                     Pattern::Bom | Pattern::Package => unreachable!(),
                 }
             })
-            .map(|(component, target_kind)| {
+            .map(|component| {
+                let target_kind = &target_kinds.0[component.bom_ref.as_ref().unwrap()];
```
Shnatsel, line 770: It's possible to only perform one hash table lookup instead of one, but it's not worth the trouble - there are only a handful of these calls, and it's not worth the additional complexity this would introduce with a `filter_map` call.

lfrancke: Agreed

lfrancke (approved):

PR #631, opened 2024-02-27
Fill in the changelog for cargo-cyclonedx v0.5.0

Shnatsel:


lfrancke:
Is it also writing CycloneDX 1.4 now?

Shnatsel:
No. Do we want it to? 

AFAIK the data it outputs are exactly the same between 1.3 and 1.4, so we can set that field to pretty much whatever, but 1.3 allows for broader compatibility.

lfrancke:
As stupid as it sounds: The German guidelines for SBOMs say that a SBOM should be 1.4 or later...

I'm only on my phone and can't link it right now. So there would be value in it.

But I'm fine with releasing as is and move to a new version with 1.5 or 1.6

Issue #632, opened 2024-02-28
`cargo cyclonedx`: make CycloneDX version configurable


Shnatsel:
There are two conflicting interests pulling the CycloneDX version in opposite directions:

1. The older the version we write is, the more software can read the data
2. Certain regulations require a more recent version (e.g. 1.4 or later in Germany)

We should add a CLI argument to specify the CycloneDX version to be written, and default to the oldest version as long as all versions contain the same data (which is currently the case).

Shnatsel:
Done in #634

Issue #605, opened 2024-01-24
`--output-prefix` conflicts with `--output-pattern` (and is not really a prefix)


Shnatsel:
`--output-prefix` flag currently defines not the prefix, but the entire filename (sans the extension). This flag conflicts with `--output-pattern`, so it is not possible to create filenames like `custom_prefix-binary_name-target_platform.cdx.json` which seems potentially useful - e.g. creating files starting with UUID and then being able to automatically discover all SBOMs generated by this run.

This is a behavior we have inherited, and it doesn't seem to make much sense. These two flags should be made compatible.

Shnatsel:
Fixed in #634

Issue #562, opened 2023-11-16
Present `cargo metadata` output to the user even when there are no errors


Shnatsel:
`cargo metadata` shows a lot of progress information as it works - e.g. "cloning crates.io index" with a progress bar, or cloning a particular repo, also with a progress bar; for large projects that have never been build on this machine before it may take several minutes, and it would be nice to inform the user of what is happening during those minutes, to make it clear the tool didn't just hang.

Shnatsel:
Fixed by #634

PR #634, opened 2024-02-28
Revamp CLI, add CycloneDX spec version configuration

### LINKED ISSUE #633: `--output-pattern` flag is confusing
Right now `cargo cyclonedx` has the `--output-pattern` flag that does two things at once, and isn't really documented as such. 

It controls both the file name and what the generated BOM describes:

| flag value   | file name | BOM describes           |
|--------------|-----------|-------------------------|
| bom          | "bom"     | package                 |
| package      | $package  | package                 |
| binary       | $target   | executables and cdylibs |
| cargo-target | $target   | all compilation targets |

And there's also a separate `--output_cdx` flag that (despite its name) controls whether the `.cdx` extension is added. It is implied by `--output-pattern` being set to anything other than `bom`.

So the filename options actually are `bom.{xml,json}`, `bom.cdx.{xml,json}` and `$package.cdx.{xml,json}` for describing packages, and `$target.cdx.{xml,json}` for all other cases.

The CLI does not communicate that and is needlessly confusing.

Shnatsel:
1. Show progress information from `cargo metadata` by default (no more "why is it taking so long?!")
2. Switch `-q` to disabling progress reports, add `-qq` to suppress warnings (previously done by `-q`)
3. Remove `--output-prefix`, `--output-pattern`, `--output-cdx` flags, subsumed by the newly introduced `--override-filename` flag
4. Split configuration of what the SBOM describes into a separate flag `--describe`
5. Make CycloneDX version configurable

Fixes #633, #605, #562, #632

### DIFF
--- a/cargo-cyclonedx/src/cli.rs
+++ b/cargo-cyclonedx/src/cli.rs
@@ -184,8 +184,8 @@ impl Args {
 
 #[derive(Error, Debug, PartialEq, Eq)]
 pub enum ArgsError {
-    #[error("Invalid prefix from CLI")]
-    CustomPrefixError(#[from] FilenameOverrideError),
+    #[error("Invalid filename")]
+    FilenameOverrideError(#[from] FilenameOverrideError),
 }
 
 #[cfg(test)]
--- a/cargo-cyclonedx/src/cli.rs
+++ b/cargo-cyclonedx/src/cli.rs
@@ -8,6 +8,7 @@ use cargo_cyclonedx::{
     platform::host_platform,
 };
 use clap::{ArgAction, ArgGroup, Parser};
+use cyclonedx_bom::models::bom::SpecVersion;
 use std::collections::HashSet;
 use std::iter::FromIterator;
 use std::path;
@@ -95,6 +96,10 @@ Defaults to the host target, as printed by 'rustc -vV'"
     /// Add license names which will not be warned about when parsing them as a SPDX expression fails
     #[clap(long = "license-accept-named", action=ArgAction::Append)]
     pub license_accept_named: Vec<String>,
+
+    /// The CycloneDX specification version to output: `1.3` or `1.4`
+    #[clap(long = "spec-version")]
+    pub spec_version: Option<SpecVersion>,
 }
 
 impl Args {
@@ -164,6 +169,7 @@ impl Args {
         });
 
         let describe = self.describe.clone();
+        let spec_version = self.spec_version.clone();
 
         Ok(SbomConfig {
             format: self.format,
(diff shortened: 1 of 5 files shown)


cargo-cyclonedx/src/config.rs
```diff
@@ -253,30 +234,55 @@ pub enum ParseMode {
     Lax,
 }
 
+/// What does the SBOM describe?
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
+pub enum Describe {
```
justahero, line 239: One possible expansion here that may help users understand these options better is to add the  `clap::ValueEnum` derive.

```rust
#[derive(clap::ValueEnum, Copy, Clone, Debug, Default, PartialEq, Eq, Serialize)]
#[serde(rename_all = "kebab-case")]
pub enum Describe {
    // ..
}
```

then the supported strings with their descriptions are displayed as well for the `--describe` option

Feel free to ignore if you think the help is already sufficient.

justahero (approved):
Looks good to me, the CLI options feel clearer now.

cargo-cyclonedx/src/cli.rs
```diff
@@ -105,6 +96,10 @@ Defaults to the host target, as printed by 'rustc -vV'"
     /// Add license names which will not be warned about when parsing them as a SPDX expression fails
     #[clap(long = "license-accept-named", action=ArgAction::Append)]
     pub license_accept_named: Vec<String>,
+
+    /// The CycloneDX specification version to output: `1.3` or `1.4`
+    #[clap(long = "spec-version")]
+    pub spec_version: Option<SpecVersion>,
```
lfrancke, line 102: Can we document the default?

cargo-cyclonedx/src/cli.rs
```diff
@@ -23,23 +25,26 @@ pub enum Opts {
...
     pub manifest_path: Option<path::PathBuf>,
 
     /// Output BOM format: json, xml
     #[clap(long = "format", short = 'f', value_name = "FORMAT")]
     pub format: Option<Format>,
 
-    /// Use verbose output (-vv very verbose/build.rs output)
+    /// What to describe in the SBOM: `crate`, `binaries` or `all-cargo-targets`
+    #[clap(long = "describe")]
+    pub describe: Option<Describe>,
```
lfrancke, line 39: We should document the default

lfrancke (changes requested):
The same minor comment twice.
I tested it and it looks good to me otherwise.

cargo-cyclonedx/src/config.rs
Shnatsel: Excellent idea! Done.

Shnatsel:
All the review comments are addressed. Merging.

PR #635, opened 2024-02-28
Deduplicate specs modules

pvdrz:
This PR introduces a new `specs::common` module which contains all the shared modules between `specs::v1_3` and `specs::v1_4`.


justahero:
Looks good so far. One concern I have is that some of the commonn models may be expanded in later releases, for example the `organizationalEntity` in [version 1.5](https://cyclonedx.org/docs/1.5/json/#components_items_supplier) received an additional field `bom-ref`. Imho it's not a big issue, this field needs to be ignored for version 1.3 & 1.4 in the serialization / validation step then.

justahero (approved):

PR #643, opened 2024-03-12
Remove unused impl

justahero:
Small clean up.


pvdrz (approved):

Shnatsel:
I believe this is semver breaking change, but we're breaking semver in the next release regardless so LGTM.

PR #647, opened 2024-03-13
Adds a method to parse a generic serde_json::Value

lfrancke:
This adds a method to be able to parse a generic json_serde::Value
This is not currently possible because the specs are private.


lfrancke:
Agreed, this can be closed. Thanks!

PR #653, opened 2024-03-25
Use `jsonschema` crate to validate JSON

justahero:
This adds the function `validate_json_with_schema` to validate a `serde_json::Value` with the related JSON schema. All required json schema files are taken from the [specification](https://github.com/CycloneDX/specification) repository.

The JSON related specification tests in `cyclonedx-bom/tests` convert valid JSON files into Strings before its snapshot is checked. Now they are also validated against the JSON schema version to check its output conforms to it.


Shnatsel (approved):
It's great to have this! Thanks!

PR #658, opened 2024-04-02
Add 'bom-ref' to organizational contact

justahero:
This adds support for the `bom-ref` attribute in the `OrganizatioalContact` type. The attribute has been added in spec version 1.5.

### DIFF
--- a/cyclonedx-bom/src/models/organization.rs
+++ b/cyclonedx-bom/src/models/organization.rs
@@ -21,7 +21,7 @@ use crate::{
         normalized_string::{validate_normalized_string, NormalizedString},
         uri::{validate_uri, Uri},
     },
-    validation::{Validate, ValidationContext, ValidationResult},
+    validation::{Validate, ValidationContext, ValidationError, ValidationResult},
 };
 
 use super::bom::{BomReference, SpecVersion};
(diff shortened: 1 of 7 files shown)


cyclonedx-bom/src/models/composition.rs
```diff
@@ -101,6 +101,21 @@ impl AggregateType {
...
+impl BomReference {
+    pub fn new<T>(input: T) -> Self
+    where
+        T: ToString,
+    {
+        Self(input.to_string())
+    }
+}
+
+impl AsRef<String> for BomReference {
```
pvdrz, line 113: I think `AsRef<str>` might be more appropriate. You'd rarely use any `String` specific method considering that you cannot mutate the underlying string through `as_ref` and it will save you at least a deref.

pvdrz (changes requested):

cyclonedx-bom/src/models/organization.rs
```diff
@@ -53,15 +55,26 @@ impl OrganizationalContact {
...
             .add_field_option("email", self.email.as_ref(), validate_normalized_string)
             .add_field_option("phone", self.phone.as_ref(), validate_normalized_string)
             .into()
     }
 }
 
+/// The `bom-ref` attribute was added in spec version 1.5
+fn validate_bom_ref(_bom_ref: &BomReference, version: SpecVersion) -> Result<(), ValidationError> {
+    if version <= SpecVersion::V1_4 {
+        return Err("Attribute 'bom-ref' not supported".into());
```
Shnatsel, line 73:
```suggestion
        return Err("Attribute 'bom-ref' not supported in this format version".into());
```

cyclonedx-bom/src/models/composition.rs
Shnatsel: Agreed. I'd like to see this addressed before I merge the PR.

Shnatsel (changes requested):
Looks good by and large, just needs two small changes

cyclonedx-bom/src/models/composition.rs
justahero: Implementation deleted, was not used so far.

Shnatsel (approved):

PR #671, opened 2024-04-12
Remove `ToString` implementations

pvdrz:
`ToString` is not supposed to be implemented manually as any type implementing `Display` will automatically implement `ToString`.

I replaced all the `ToString` implementations and used the `strum` crate when possible.


Shnatsel:
That was error-prone boilerplate and I am glad to be rid of it. Thanks!

PR #683, opened 2024-04-29
Expand `Vulnerability` type to support spec version 15

justahero:
This expands the `Vulnerability` & associated types to add fields to support spec version 1.5.

* add fields `workaround`, `proof_of_concept` & `rejected` to `Vulnerability` type
* add `ProofOfConcept` type
* add fields `first_issued` & `last_updated` fields to `VulnerabilityAnalysus` type


Shnatsel (approved):
Looks good, just gotta appease DCO

PR #705, opened 2024-05-15
feat: allow using a pre-parsed JSON

ctron:
This changes adds the ability to use a preparse JSON for the 1.4 and "detector" method too.

### DIFF
--- a/cyclonedx-bom/src/models/bom.rs
+++ b/cyclonedx-bom/src/models/bom.rs
@@ -128,8 +128,12 @@ impl Bom {
     pub fn parse_from_json<R: std::io::Read>(
         mut reader: R,
     ) -> Result<Self, crate::errors::JsonReadError> {
-        let json: serde_json::Value = serde_json::from_reader(&mut reader)?;
+        Self::parse_json_value(serde_json::from_reader(&mut reader)?)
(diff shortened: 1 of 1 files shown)


cyclonedx-bom/src/models/bom.rs
```diff
@@ -128,8 +128,12 @@ impl Bom {
     pub fn parse_from_json<R: std::io::Read>(
         mut reader: R,
     ) -> Result<Self, crate::errors::JsonReadError> {
-        let json: serde_json::Value = serde_json::from_reader(&mut reader)?;
+        Self::parse_json_value_value(serde_json::from_reader(&mut reader)?)
+    }
 
+    /// General function to parse a pre-parsed JSON file, fetches the `specVersion` field first,
+    /// then applies the right conversion.
+    pub fn parse_json_value_value(json: Value) -> Result<Self, crate::errors::JsonReadError> {
```
Shnatsel, line 136: I like the basic direction. But shouldn't this be called `parse_json_value` rather than `parse_json_value_value`?

ctron: Doppelmoppel :grimacing: … fixed it.

PR #707, opened 2024-05-16
chore: allow turning a normalized string into a string

ctron:
Having a `NormalizedString` and wanting to consume it as a `String` isn't possible right now without cloning the `String`. I think this is not necessary. This PR adds a way to turn it back into a `String`.


lfrancke (approved):
That looks sane to me but I'm a bit out of touch.
I'll let it here for the others to look at for a day or two but if I don't hear anything else I can merge it later.

Thank you!

PR #708, opened 2024-05-16
chore: allow parsing the timestamp as datetime without cloning it

ctron:
Having a `DateTime` (which is actually a `String`), I would like to parse it into an actual `OffsetDateTime`. For that I only need a `&str`, however I can only get a clone of `DateTime`'s string.

This PR adds `AsRef<str>` for `DateTime`, so that it is possible working with the string data.

### DIFF
--- a/cyclonedx-bom/src/external_models/date_time.rs
+++ b/cyclonedx-bom/src/external_models/date_time.rs
@@ -77,6 +77,12 @@ impl std::fmt::Display for DateTime {
     }
 }
 
+impl AsRef<str> for DateTime {
+    fn as_ref(&self) -> &str {
+        &self.0
+    }
+}
+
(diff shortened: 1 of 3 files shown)


lfrancke (approved):
Same as the other one.

Thank you! I'll let it here for a few days.

ctron:
@lfrancke maybe you can take another look. I found the same limitation in two more places. I assumed it makes sense to amend this PR instead of creating another one.

ctron:
Found one more, and I found a case (`Cpe`) where it seems impossible to get to the actual value.

lfrancke (approved):

lfrancke:
Thanks for finding these! Feel free to add more to this until we merge it :)

PR #710, opened 2024-05-21
Ship `cyclonedx-bom` v0.6.0

Shnatsel:


Shnatsel:
The CI failure seems to be due to a bug in the latest Rust nightly. The `time` crate fails to compile under it.

Shnatsel:
It is already reported upstream and only affects nightly builds: https://github.com/rust-lang/rust/issues/125319

So I don't think this should be a release blocker.

PR #711, opened 2024-05-21
Fill in description in `cyclonedx-bom-macros`, which is mandatory for publishing to crates.io

Shnatsel:


Shnatsel:
CI failing due to a regression in rustc nightly, will be fixed by #710

PR #723, opened 2024-06-03
fix: process detected 1.5 spec version

ctron:
Spec version 1.5 is detected, but then rejected as "not supported". Adding the version and dropping the "catch all" should do the trick.


Shnatsel:
Thanks!

We didn't really test for this since enum matching was supposed to be exhaustive, but the catch-all messed with things. I'll merge as soon as CI passes, and try to ship a new version with the fix in the next few days.

PR #719, opened 2024-05-25
Remove `--allow-dirty` option (v2) …

Olf0:
… from `cargo publish` calls in the CI/CD workflows, because this option prevents creating a `.cargo_vcs_info.json` file.  This should (really) close https://github.com/rust-lang/crates.io/issues/8551, see there for details.
Now a `cargo package` with `--allow-dirty` executes all steps a `cargo publish` does except publishing, then the changes are committed back to the git repo and lastly `cargo publish` is run; for details see last paragraph of https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/702#issuecomment-2130606809.

Supersedes PR #702

Signed-off-by: olf <Olf0@users.noreply.github.com>

PR #724, opened 2024-06-03
Rewrite and simplify crates.io publishing actions

Shnatsel:
This changes the workflow to be triggered by a tag, rather than create a tag automatically. We need the tag to be created by a human rather than by Github Actions so that cargo-dist workflows that publish binaries would work properly. (I had to do some manual fixup to build binaries for the latest release).

This also removes the `--allow-dirty` flag that prevents attribution to a specific commit and is known to cause issues for links in the way our documentation shows up on crates.io, previously attempted in #719

Drops cache because the check before publishing uses latest versions from crates.io rather than what we have in Cargo.lock, so caching isn't do much there. Also, clean builds have less of a risk of running into incremental compilation bugs in the compiler, and we run this workflow rarely enough that execution time is not a concern.

### DIFF
--- a/.github/workflows/deploy_cargo_cyclonedx.yml
+++ b/.github/workflows/deploy_cargo_cyclonedx.yml
@@ -1,41 +1,15 @@
 name: Deploy cargo-cyclonedx to Crates.io
 
 on:
-  workflow_dispatch:
-    inputs:
-      releaseType:
-        description: "cargo-cyclonedx release type (major, minor, patch)"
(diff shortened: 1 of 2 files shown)


Shnatsel:
I've tested this on a fork but it didn't have a token so I can only be sure this triggers correctly on a tag, and not about any of the more involved stuff. But it's still `cargo publish` and we aren't making any untracked files in the package directory so it should work fine.

PR #731, opened 2024-06-14
chore(dist): update to 0.16 and enable github artifact attestations

ashleygwilliams:


.github/workflows/release.yml
```diff
@@ -38,15 +40,15 @@ permissions:
...
   push:
     tags:
       - '**[0-9]+.[0-9]+.[0-9]+*'
-  pull_request:
 
 jobs:
   # Run 'cargo dist plan' (or host) to determine what tasks we need to do
   plan:
-    runs-on: ubuntu-latest
+    runs-on: "ubuntu-20.04"
```
Shnatsel, line 51: I understand this is the oldest still-supported version outside of paid subscriptions, in order to have a reasonably low glibc version requirement?

I wish there was an alias for that, instead of requiring us to manually bump this once this version stops receiving security updates. But I'm not going to block this PR on it.

Shnatsel (approved):

Shnatsel:
Thank you!

.github/workflows/release.yml
ashleygwilliams: yeah i hear ya, it's something the axo team pays a lot of attention to and so when it's close to EOL we'll release a new version well ahead of time. but you're not wrong that it's unfortunate that there's no abstraction (that i'm aware of)

Shnatsel: According to the docs, there isn't: https://docs.github.com/en/actions/using-jobs/choosing-the-runner-for-a-job

Also `ubuntu-latest` uses 22.04, not 24.04, so that name is a lie.

ashleygwilliams: lol, fair enough!

PR #732, opened 2024-06-24
Fix tools type in metadata

justahero:
The `tools` type in `metadata` allows optional `services` & `components` fields according the [1.5 specification](https://cyclonedx.org/docs/1.5/json/#metadata_tools_oneOf_i0_components). These are not required therefore these fields have changed to `Option` fields.

Previously this has not been covered by tests.

### DIFF
--- a/cyclonedx-bom/src/models/tool.rs
+++ b/cyclonedx-bom/src/models/tool.rs
@@ -36,8 +36,8 @@ pub enum Tools {
 
     /// Added in 1.5
     Object {
-        services: Services,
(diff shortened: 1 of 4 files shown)


lfrancke (approved):
Looking good to me. I'll approve but we can give others a chance as well.
I tested it on my failing BOM and it works now.

I'm wondering what'll happen for items that contain neither services nor components. It's probably still valid either way.

You could update the changelog but I believe @Shnatsel has always done it at release time

PR #740, opened 2024-07-09
chore: update dependencies to newer versions

ctron:
Mainly to update jsonschema -> reqwest -> hyper to a 1.x version.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -55,7 +55,7 @@ jobs:
             continue-on-error: false
           - # test with the MSRV version
             test-arm: "Minimum Supported Rust Version"
-            rust-version: "1.70.0"
+            rust-version: "1.74.0"
             continue-on-error: false
           - # test with the current nightly version (allowed to fail)
             test-arm: "Nightly"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,21 +3,6 @@
 version = 3
 
 [[package]]
-name = "addr2line"
-version = "0.21.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
-dependencies = [
- "gimli",
-]
-
-[[package]]
-name = "adler"
-version = "1.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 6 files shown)


Shnatsel:
I think a better idea is to just disable the `resolve*` features that pull in `hyper`. We really don't want to pull in the entirety of reqwest, hyper and tokio unconditionally from a serialization/deserialization crate.

Also that's definitely a no on the `packageurl` upgrade - that's certainly semver-breaking, and if we're breaking semver we should just switch to the `purl` crate which is much better all around.

ctron:
> I think a better idea is to just disable the `resolve*` features that pull in `hyper`. We really don't want to pull in the entirety of reqwest, hyper and tokio unconditionally from a serialization/deserialization crate.
> 
> Also that's definitely a no on the `packageurl` upgrade - that's certainly semver-breaking, and if we're breaking semver we should just switch to the `purl` crate which is much better all around.

Ok, I'll try to update the PR tomorrow.

Shnatsel:
The `jsonschema` crate pulls in a lot of dependencies and shouldn't be a runtime dependency at all. We should turn it into a dev-dependency. There is now an issue about it: https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/741

ctron:
Ok, I changed it to the following:

* packageurl is back to 0.3
* the jsonschema is an optional dependency, enabled by default. As the validate method is a public method, that would break the API. This can be turned into off-by-default with the next breaking release.
* Also, jsonschema has `default-features=false`, which is sufficient for this crate

Shnatsel:
Turning `jsonschema` into an optional feature is a breaking change if someone was using the crate with `default-features = false`, but I don't think anyone has been doing that on account of the crate not having any features.

The Nix flake CI is failing due to increased MSRV. It expects 1.70, while this PR bumps it to 1.74. 1.74 is very recent, and I think we would like to stick to 1.70 for now.

What is the motivation for the upgrades? We could consider bumping it if it's *really* worth it.

Shnatsel:
Since this appears to have stalled, I started pruning the dependency tree myself. See #744 and #746

ctron:
> Since this appears to have stalled, I started pruning the dependency tree myself. See #744 and #746

Sorry for that. Yes, that might be quicker.

Shnatsel:
We've dropped `reqwest` entirely in #744 and #750, so I'm going to go ahead and close this.

PR #748, opened 2024-07-17
Drop OpenSSL dependency from Nix flake

Shnatsel:
Now that we no longer depend on `cargo` as a library which in turn depends on OpenSSL, we don't need OpenSSL as dependency anywhere.

I don't know Nix, so...

712be92c4bd4a36bbc01d63b7e431770ed65615201ea3c12e541b84b024edf45

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -52,7 +52,6 @@
           pname = "cyclonedx-rust-cargo";
           version = "0.1.0";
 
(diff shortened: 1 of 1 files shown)


Shnatsel:
Well, it seems to have worked, and we don't have any reviewers who know Nix, so I'm going to go ahead and merge this.

PR #750, opened 2024-07-17
Use `jsonschema` only as a dev-dependency

### LINKED ISSUE #741: Move JSON schema validation into `#[test]` scope
Move the function [`validate_json_with_schema`](https://github.com/CycloneDX/cyclonedx-rust-cargo/blob/main/cyclonedx-bom/src/schema.rs) into the `test` scope. This is to limit a lot of dependencies to tests that are pulled in by the `jsonschema` crate.

Users of the `cyclonedx-bom` should not require these dependencies, the generated JSON BOM files in `cyclonedx-bom` already conform to their schemas.

Shnatsel:
Depends on #749

Fixes #741 

Also add `exclude` directives to `Cargo.toml` to cut the tarball size in half.

You'd think you could just stick `#[cfg(test)]` on a module and call it a day, but `#[cfg(test)]` isn't actually set for the main crate when running integration tests. Making a separate crate for the test utils is the least clunky option: https://stackoverflow.com/a/44545091/585725


Shnatsel:
schema.rs was just moved from the main crate to the utils crate with minimal modifications, I don't know why Github presents the diff in such a clunky way.

justahero (approved):

Issue #628, opened 2024-02-24
`cyclonedx-bom` is unsuitable for `cargo auditable` use case


Shnatsel:
There are several issues that are interfering with using `cyclonedx-bom` to write CycloneDX in `cargo auditable`

`cargo auditable` operates under very different constraints to regular SBOMs; this is what drove me to use a custom format as opposed to using an off-the-shelf CycloneDX or SPDX. The issues currently affecting it are:

 - [ ] Toplevel `version` has been made a required field in #618 and is always written. It needs to be omitted to save bytes.
 - [ ] Toplevel `serialNumber` is always written, even if set to `None`. It needs to be omitted to save bytes.
 - [ ] There is no way to write a minified JSON. `output_as_json_v1_3` and `output_as_json_v1_4` pretty-print the JSON, wasting space. The `spec` module required to implement it outside `cyclonedx-bom` is private.

PR #751, opened 2024-07-17
Do not pretty-print JSON

Shnatsel:
Do not include whitespace in the generated JSON by default. This saves storage and network bandwidth.

Humans that want to inspect it can pipe it to `jq`, open it in the browser to get a nice UI to explore it, or use one of the many other tools for JSON inspection and manipulation.

Addresses one of the items from #628


Shnatsel:
Hm, this breaks all the snapshot tests. I guess I'll have to approach it in a different way and make pretty-printing configurable.

PR #753, opened 2024-07-17
Drop the in-tree nix flake

Shnatsel:
There is a Nix flake in the proper Nix repositories now: https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/tools/rust/cargo-cyclonedx/default.nix

So we do not need to keep the unmaintained in-tree one.

Packaging status across various distributions, by the way: https://repology.org/project/cargo-cyclonedx/versions


lfrancke:
Should `.envrc` also go?

Shnatsel:
Yep. Good catch!

PR #755, opened 2024-07-18
more explicit build dependency handling

thillux:
This is a follow up on #736 which was merged and later reverted because of a infinite loop I accidentally produced there.

This PR differs in the following:
- test for cyclic dependencies added
- fixed dependency kind indexing under presence of cyclic dependencies
- changed name of command line option for no build deps in SBOM, as dev deps are already excluded


cargo-cyclonedx/src/generator.rs
```diff
@@ -542,6 +593,45 @@ fn index_resolve(packages: Vec<Node>) -> ResolveMap {
...
+            let dep_kind = strongest_dep_kind(child_dep.dep_kinds.as_slice());
+            let child_node_kind = min(parent_node_kind, dep_kind);
+
+            let dep_kind_on_previous_visit = id_to_dep_kind.get(&child_node.id);
+            // insert/update a nodes dependency kind, when its new or stronger than the previous value
+            if dep_kind_on_previous_visit.is_none()
+                || child_node_kind > *dep_kind_on_previous_visit.unwrap()
+            {
+                let _ = id_to_dep_kind.insert(&child_node.id, child_node_kind);
+                nodes_to_visit.push_front(child_node);
```
Shnatsel, line 624: While this doesn't loop infinitely anymore, it would fail to index some dependencies.

Consider the following scenario:

There is a runtime dependency on `rand` with `default-features = false`, and a build dependency on `rand` with default features enabled. In this setup `rand_chacha` crate is only present as a build dependency, not a normal one.

If this code indexes the runtime `rand` dependency first, then the condition `child_node_kind > *dep_kind_on_previous_visit.unwrap()` on the build dependency will fail, and the children of this node will not be traversed. Therefore we will never reach and index `rand_chacha`. It will then be erroneously reported as a runtime dependency because of ` .unwrap_or(PrivateDepKind::Runtime)` on line 106.

This bug is particularly insidious because it is dependent on the iteration order. If we reach the build-dependency on `rand` first, then everything will work out fine. But if the normal dependency on `rand` is reached first, some build dependencies will not be indexed.

This would also work fine on older versions on Cargo or even on older Rust editions that do not enable Cargo resolver v2. With resolver v2 there are different feature sets for build and runtime dependencies, which makes it possible for `rand` to have different dependencies in normal and build-only contexts.

I think the proper way to fix this is to write an honest DFS that keeps track of the elements already visited in a set (HashSet or BTreeSet). And we need to keep track of `(Node, PrivateDepKind)` in the visited set, not just `Node`. In this case `DependencyKindMap` is not enough because it only keeps track of the strongest dependency kind, while we need to keep traversing the tree whenever we see a new dep kind we haven't seen before for this package.

cargo-cyclonedx/src/generator.rs
```diff
@@ -542,6 +593,45 @@ fn index_resolve(packages: Vec<Node>) -> ResolveMap {
         .collect()
 }
 
+fn index_dep_kinds(root: &PackageId, resolve: &ResolveMap) -> DependencyKindMap {
+    // cache strongest found dependency kind for every node
+    let mut id_to_dep_kind: HashMap<&PackageId, PrivateDepKind> = HashMap::new();
+    id_to_dep_kind.insert(root, PrivateDepKind::Runtime);
+
+    let root_node = &resolve[root];
+    let mut nodes_to_visit = LinkedList::<&Node>::new();
```
Shnatsel, line 602: Probably not relevant for performance in the grand scheme of things, but still: why a linked list? It is almost never the right tool for the job. Wouldn't a simple `Vec` where you `push` and `pop` be enough? That's semantically equivalent to `push_front` + `pop_front` on a linked list.

Shnatsel (changes requested):
I believe I have found a bug, please check the review comments.

cargo-cyclonedx/src/generator.rs
thillux: Good point. Will change to a `Vec`.

cargo-cyclonedx/src/generator.rs
thillux: Good point. I tested with rand as you described on my branch. I currently does not work, but neither does `cargo auditable`, as features are not taken into account correctly in the generator code currently.

My suggestion would be, that I cleanup this PR and start another PR that traverses the features, which is also missing in the other generator code, which does not deal with my indexing stuff.

Full disclosure: I cannot promise, when I have time for this next PR ATM. I hope next week or so.

Shnatsel: I've written `cargo auditable` before Resolver v2 was introduced, so I guess that makes sense. I'll take another look at that.

> My suggestion would be, that I cleanup this PR and start another PR that traverses the features, 

I'm not sure what you mean by "traversing features", could you elaborate?

Shnatsel: I've confirmed that `cargo auditable` also has this bug. Here's a test fixture that reproduces it completely offline: https://github.com/rust-secure-code/cargo-auditable/commit/521e4f99cae86400df4da31e0b6391f063e40d77

Making tests runnable offline is important for some Linux distributions.

Shnatsel: I've done some digging, and it turned out that `cargo metadata` [does not support Resolver v2 either](https://rust-lang.github.io/rfcs/2957-cargo-features2.html#cargo-metadata). While the bug is real, there is no way for us to fix it given the data `cargo metadata` provides. And there is no viable alternative data source at this time. We would need https://github.com/rust-lang/rfcs/pull/3553 for a proper fix.

So we'll just have to make do with what we have.

Shnatsel:
I retract what I said about resolver v2 - we cannot fix that edge case because `cargo metadata` doesn't support resolver v2 either.

Please fix the test failures and I'll be happy to merge this.

thillux:
@Shnatsel ready & thanks for your feedback regarding resolver v2.

Shnatsel:
Thank you, I will take a look in the next few days.

My revised feedback about resolver v2 is to ignore its existence. `cargo metadata` does not support it, so there is no way for us to support it either.

I am really sorry about sending you on this wild goose chase. I forgot that Cargo just does not expose enough data for us to handle this correctly.

Shnatsel:
I'm going to go ahead and merge this to expedite the process. What nits I have I'll address myself in a follow-up PR.

Thank you!

PR #758, opened 2024-08-06
Change visibility of model types

### LINKED ISSUE #739: cyclonedx_bom: make Cpe constructable
The current implementation of Cpe in https://github.com/CycloneDX/cyclonedx-rust-cargo/blob/main/cyclonedx-bom/src/models/component.rs does not allow access to the member variable and provides no constructors. Some way to construct Cpes in applications which use the bom library should be provided.

eg.:
`let cpe = Cpe(cpe_string);`

Result:

```
note: constructor is not visible here due to private fields
   --> /home/scott/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cyclonedx-bom-0.4.3/src/models/component.rs:439:16
    |
439 | pub struct Cpe(pub(crate) String);
    |                ^^^^^^^^^^^^^^^^^ private field
```

    
We can implement, if you let us know if you prefer public member variable, a `new `constructor or conversion from `String `or `&str`, etc. Purl, for instance implements both `new `and `FromStr`

justahero:
Instead of setting the visibility of model types to the crate level, types are public now. When building the BOM some of the models were not able to be constructed. With the change inner BOM model types can now be constructed / used by external crates directly.

Instead of only adapting the `Cpe` type (as given in issue #739) all types in the `/models` folder are public now, including their associated validation functions.

Closes #739

### DIFF
--- a/cyclonedx-bom/src/models/annotation.rs
+++ b/cyclonedx-bom/src/models/annotation.rs
@@ -45,12 +45,12 @@ impl Validate for Annotations {
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Annotation {
-    pub(crate) bom_ref: Option<String>,
-    pub(crate) subjects: Vec<String>,
-    pub(crate) annotator: Annotator,
-    pub(crate) timestamp: DateTime,
-    pub(crate) text: String,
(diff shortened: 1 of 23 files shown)


ModestMannfred:
Awesome! Thank you!

PR #761, opened 2024-08-07
Support parsing of empty XML string tags

### LINKED ISSUE #738: XML deserialization of empty tags is incorrect
As a working example, I am going to reference the following area of code, but I believe it is possible to run into this in any area that the library is parsing an `Option<String>` field: https://github.com/CycloneDX/cyclonedx-rust-cargo/blob/2911287b2520a7ddab1782b48c35112279b1be17/cyclonedx-bom/src/specs/common/component.rs#L513-L517

Setting context, I was attempting to swap out an ad-hoc CycloneDX parser with `cyclonedx-bom`. One of the tests for my ad-hoc parser was to ensure that the keycloak BOM found here loads: https://github.com/CycloneDX/bom-examples/blob/c0436d86cd60693f01d19fe1aacfd01e70e17036/SBOM/keycloak-10.0.2/bom.xml. Now, it obviously doesn't because it is a v1.2 spec and `cyclonedx-bom` only supports v1.3-v1.5. But, looking through the spec and looking at the BOM, it seemed like it should load as a v1.3, so I changed the version locally. That seemed a reasonable test given that a diff between https://github.com/CycloneDX/bom-examples/blob/c0436d86cd60693f01d19fe1aacfd01e70e17036/SBOM/laravel-7.12.0/bom.1.2.xml and https://github.com/CycloneDX/bom-examples/blob/c0436d86cd60693f01d19fe1aacfd01e70e17036/SBOM/laravel-7.12.0/bom.1.3.xml only added some properties and changed the spec version. 

However, when I attempted to load the modified keycloak BOM, I got the following error:

```
    Got unexpected XML element when reading {http://cyclonedx.org/schema/bom/1.3}description: Got unexpected element EndElement({http://cyclonedx.org/schema/bom/1.3}description)
```

After digging into the issue, what I discovered was that the keycloak BOM has several empty description tags in it:

```xml
<description />
```

A quick modification of the `laravel-7.12.0/bom.1.3.xml` file to modify one of the description tags the same way caused the same error to occur. I also attempted to just modify one of the description like the following and received the same error:

```xml
<description></description>
```

The [v1.3 (XML)](https://cyclonedx.org/docs/1.3/xml/#type_component) and [v1.2 (XML)](https://cyclonedx.org/docs/1.2/xml/#type_component) specs 
both define the description field as `<bom:description> xs:normalizedString </bom:description> [0..1]` and an `xs:normalizedString` _can_ be empty.

Again, there are multiple places just in the `Component::read_xml_element` where this could be an issue. My guess is that, since the `pom.xml` file that is used to generate the CycloneDX Maven BOMs can have empty strings for the description, that is why it is getting forwarded on to the BOM that way instead of stripping out the description tag completely. 

I could make an argument for handling it either by making an empty string become a `None` for the `Option<String>` OR by literally making it an `Some(String::new())`. I think that the latter is closer to the behavior that would happen with the JSON implementation of the spec, since an equivalent there would be:

```json
"description": ""
```

and that would be parsed as `Some(String::new())` by `serde`.

justahero:
This adds support to parse either

```xml
<description>something</description>
```

or an empty XML tag.

```xml
<description />
```

The latter XML tag is an empty string (equivalent to `<description></description>`).

* fix `read_simple_tag` function to allow to parse empty XML tags
* add documentation
* add test sample with an empty `<description>` tag

Closes #738

### DIFF
--- a/cyclonedx-bom/src/xml.rs
+++ b/cyclonedx-bom/src/xml.rs
@@ -442,6 +442,17 @@ impl FromXmlType for f32 {
     }
 }
 
+/// Reads a simple String tag.
+///
+/// ```xml
+/// <description>Content</description>
+/// ```
+/// &
+/// ```xml
+/// <description />
+/// ```
+///
+/// are valid XML tags. The first returns the string "Content", the latter is an empty string.
 pub(crate) fn read_simple_tag<R: Read>(
     event_reader: &mut EventReader<R>,
     element: &OwnedName,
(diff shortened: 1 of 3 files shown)


Shnatsel (approved):

Shnatsel:
Thank you!

PR #770, opened 2024-11-22
Include author for components when available in Cargo metadata

davids-work:


Shnatsel:
I see v1.6 has [deprecated the `author` field](https://cyclonedx.org/docs/1.6/json/#components_items_author) and added `authors` which allows multiple entries. However it is not available in v1.5 or earlier.

@davids-work what is the motivation for listing the authors like this? Do you need to meet some regulations that require the authors to be listed?

davids-work:
Indeed, or at least tthe internal interpretation of the regulation(s) that we need to fulfill is that we must include the author field for every dependency.

I don't think that there are any rules for the exact formatting (comma-separated or otherwise) in case of multiple authors. I note that the v1.4 spec says "The person(s) or organization(s) that authored the component", so the spec writers seems to at least have had the notion that there might be multiple authors involved.

PR #774, opened 2024-12-13
Upgrade to dist v0.26 to build Aarch64 Linux binaries

Shnatsel:
This version of dist also adds `cargo cyclonedx` support to generate SBOMs, but this PR doesn't take advantage of that yet.

cc @lfrancke - I remember you were interested in aarch64 Linux binaries


lfrancke:
Happy to take a look at it if you want a review but I won't get to it this week anymore :(

Shnatsel:
This is just FYI. I think this change is too trivial to require review (or if it breaks something, we won't see it coming from just the diff) so I'm going to go ahead and merge it myself.

PR #777, opened 2025-02-11
feat: use source as vcs url

barblin:


barblin:
@Shnatsel, what do you think?

barblin:
> @Shnatsel, what do you think?

@justahero ?

Shnatsel:
Heya, sorry this slipped through the cracks. Could you explain the motivation behind this change? What does this accomplish?

barblin:
> Heya, sorry this slipped through the cracks. Could you explain the motivation behind this change? What does this accomplish?

@Shnatsel attempts to add the repo url as a VCS ExternalReference, to enrich the component information and link it to an origin repository

barblin:
> According to https://doc.rust-lang.org/cargo/commands/cargo-metadata.html#json-format this field is meant to be opaque and its content is not stable between versions and does not conform to any particular format:
> 
> > ```
> >         /* The source ID of the package, an "opaque" identifier representing
> >            where a package is retrieved from. See "Compatibility" above for
> >            the stability guarantee.
> > 
> >            This is null for path dependencies and workspace members.
> > 
> >            For other dependencies, it is a string with the format:
> >            - "registry+URL" for registry-based dependencies.
> >              Example: "registry+https://github.com/rust-lang/crates.io-index"
> >            - "git+URL" for git-based dependencies.
> >              Example: "git+https://github.com/rust-lang/cargo?rev=5e85ba14aaa20f8133863373404cb0af69eeef2c#5e85ba14aaa20f8133863373404cb0af69eeef2c"
> >            - "sparse+URL" for dependencies from a sparse registry
> >              Example: "sparse+https://my-sparse-registry.org"
> > 
> >            The value after the `+` is not explicitly defined, and may change
> >            between versions of Cargo and may not directly correlate to other
> >            things, such as registry definitions in a config file. New source
> >            kinds may be added in the future which will have different `+`
> >            prefixed identifiers.
> >         */
> > ```
> 
> I'm not sure it's a good idea to rely on parsing an explicitly opaque value.

Thats why we don't extract the repository url if the field does not exist

barblin:
@Shnatsel what do you think?

Shnatsel:
@beer4code thanks for the info! Let's not wait on the original author. Would you go ahead and open a new PR with your version?

beer4code:
Thanks for the fast response. Yes, can do. I'll take care of it in the next few days 👍

PR #784, opened 2025-04-02
Fix annotation serialization issue

### LINKED ISSUE #783: `Annotation` bom_ref field incorrectly serialized to JSON
The `Annotation`.`bom_ref` field gets generated as "bomRef" and should be "bom-ref". As-is it fails validation when generating annotations for 1.5.

jacobcalvert:
This PR fixes the bom-ref serialization issue in annotations. 

Closes #783


Shnatsel:
Ah, good catch! Thank you for the PR!

CI is failing, could you take a look? Perhaps some more tests need to be updated.

jacobcalvert:
Try this one. Sorry, I must have fixed one insta and not the other.  This recent commit should fix it up.

Shnatsel:
Thanks!

PR #810, opened 2025-12-06
Update a snapshot in cyclonedx-bom-macros for Rust 1.91

musicinmybrain:
Based on testing in Fedora with Rust 1.91.1, while packaging this as a new dependency for https://github.com/astral-sh/uv.


musicinmybrain:
> @musicinmybrain Is this something you'd like to provide a PR for? (no pressure, just checking)

Probably not: since sensitivity to precise compiler output is an idea that is fundamental to `trybuild` as I understand it, I’m not sure what to suggest.

PR #846, opened 2026-03-13
Prepare a release for cyclonedx-bom 0.8.1

grahamc:
After #844 and #845 I'd be glad for a release. I thought submitting the (small) changelog might be a help there :).

As I mentioned in a comment on #845, I already pinned my project to a revision. That means this isn't critical to me if you're not ready to ship a release. But I wouldn't mind the tidyness.

Thank you again!


Shnatsel:
I understand we need the semver bump because spdx crate is present in the public API? I can never remember which crates we expose and which we keep hidden.

PR #858, opened 2026-03-18
Add `--ignore-hakari` flag to prune spurious workspace-hack deps

plaidfinch:
[`cargo-hakari`](https://crates.io/crates/cargo-hakari) generates a workspace-hack crate that unifies dependency features for build performance. This crate is always an empty stub that contributes no code to any binary, but it appears in cargo metadata as a normal dependency of every crate in the workspace, and whose dependencies are every crate outside the workspace. The result of this is that *every* crate in the workspace is made to depend transitively on each external dependency of *any* crate in the workspace. Particularly problematically, this behavior causes dependencies even of test-only crates in the workspace to appear as transitive dependencies of *every* crate in the workspace, effectively pulling many dev-dependencies spuriously into the production dependency graph, even if they are only in fact used in tests.

The new `--ignore-hakari` flag detects the hakari workspace-hack crate (if it is present) by reading `.config/hakari.toml`, removes it from the dependency graph, and prunes any transitive dependencies that have hence become unreachable. This is a post-resolution transformation that does not alter feature sets on surviving components.

The test fixture was generated by cargo-hakari and reproduces the real-world scenario of "proptest shouldn't appear in our production dependencies" that prompted the idea for this feature.


plaidfinch:
Alternatively, we could add a more-general `--omit-package` flag which elides the `hakari` detection and requires manual annotation of the name of your workspace-hack crate, if applicable. This would be a trivial change to the PR and I'd be happy to take it that direction if you think that'd be better.

lfrancke:
Sorry for the long delay.

Personally I'd be more in favor of a more generic feature over a specific one. @Shnatsel any opinions?

Shnatsel:
I'm not thrilled about the extra complexity we'd have to maintain.

More importantly, since cargo-hakari alters the dependency resolution (by design), AFAIK there just isn't a way to correctly report dependency trees after cargo-hakari unified all the features in the workspace. So there just isn't a way for us to accurately construct a SBOM in presence of the workspace hack.

I think the only way to make this work correctly is to have cargo-hakari temporarily undo the hack. Even ignoring the effort of supporting this in hakari once (vs teaching every other ecosystem tool about hakari), I don't think it's technically possible to undo its the feature unification on our side without outright reimplementing Cargo's dependency resolution inside cargo-cyclonedx.

plaidfinch:
Apologies for the late reply here; that makes sense! Thanks for following up on this.

PR #861, opened 2026-03-19
Extract VCS URL from package.id (pkgid spec) instead of package.source

beer4code:
## Summary

Replace the use of `package.source` (opaque/unstable) with `package.id` (stable) for extracting VCS URLs from git dependencies.

`package.source` is explicitly documented as opaque and unstable by the cargo_metadata crate. `package.id` (the pkgid spec) is the stable alternative. On Rust 1.77+ the pkgid format for git deps is:

    git+proto://host/path[?query]#name@version

The fragment contains `name@version` (not a commit hash like `package.source` does), so the output no longer includes commit hashes.

This PR is supposed to replace #777.

## Changes

- Add `extract_git_url_from_id` helper in `purl.rs` that strips the `#name@version` fragment and `?branch=`/`?tag=`/`?rev=` query params from the pkgid, returning a clean `git+proto://host/path` URL 
- Use the helper in `get_purl` for the PURL `vcs_url` qualifier, replacing the previous `source_to_vcs_url` function
- Use the helper in `get_external_references` as a fallback when `package.repository` is absent on git dependencies
- Update `git_package.json` fixture to use the new pkgid format

### DIFF
--- a/cargo-cyclonedx/src/generator.rs
+++ b/cargo-cyclonedx/src/generator.rs
@@ -23,7 +23,7 @@ use crate::config::PlatformSuffix;
 use crate::config::SbomConfig;
 use crate::config::{IncludedDependencies, ParseMode};
 use crate::format::Format;
-use crate::purl::{extract_git_url_from_id, get_purl};
+use crate::purl::{get_purl, strip_git_url};
 
 use cargo_metadata;
 use cargo_metadata::DependencyKind;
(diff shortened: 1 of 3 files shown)


Shnatsel:
Thanks for the PR!

> Update git_package.json fixture to use the new pkgid format

I'd rather have tests and fixtures for both formats. The old for fallback to source on older versions as described in https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/854 and the new to extract it from pkgid.

beer4code:
Would you mind having another look please?

PR #812, opened 2025-12-11
Update several dependencies across SemVer boundaries

musicinmybrain:
This includes (applies on top of) https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/687.

No source-code changes seemed to be necessary *except for `spdx`*, https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/812#issuecomment-3650687508. Release notes are linked below for information on breaking changes.

- https://github.com/marshallpierce/rust-base64/blob/v0.22.0/RELEASE-NOTES.md
- https://github.com/reem/rust-ordered-float/releases/tag/v5.0.0
- https://github.com/Peternator7/strum/blob/v0.27.0/CHANGELOG.md
- https://github.com/dtolnay/thiserror/releases/tag/2.0.0
- https://github.com/EmbarkStudios/spdx/blob/0.13.2/CHANGELOG.md

PR #862, opened 2026-03-30
Update a few cargo-cyclonedx dependencies across SemVer boundaries.

musicinmybrain:
Similar to https://github.com/CycloneDX/cyclonedx-rust-cargo/pull/812 for cyclonedx-bom, this updates a few dependencies across SemVer boundaries. I’m not aware of any MSRV issues here, and `cargo test` still passes.

There are cases where it might make sense to update `Cargo.lock` due to bug or security fixes, or just to keep up with the ecosystem, but for this PR I was focused only on crossing SemVer boundaries.


lfrancke:
Hi @musicinmybrain I'm happy to merge dependency updates. Do you want to update this PR first? And did you check whether the MSRV still fits?

musicinmybrain:
> Hi @musicinmybrain I'm happy to merge dependency updates. Do you want to update this PR first?

Thanks. I think this is probably still in good shape. It looks like it’s still current with `main`, and I don’t think there are any other dependencies that need SemVer bumps to be compatible with their latest releases. I don’t personally have much interest in contributing patch-release bumps in `Cargo.lock`.

> And did you check whether the MSRV still fits?

Yes, I checked MSRV manually: https://crates.io/crates/cargo-lock/11.0.1 wants 1.85.0, https://crates.io/crates/env_logger/0.11.10 wants 1.71.0, and https://crates.io/crates/validator/0.20.0 wants 1.81.0.  All are consistent with a7f47a682409094d1708874d7c927309801e2c59.

Shnatsel:
The bumps for cyclonedx-bom have to be carefully reviewed to see if we're upgrading something that's exposed publicly, in which case the change is semver-breaking. For cargo-cyclonedx we can just merge these as long as MSRV is preserved.

Issue #802, opened 2025-09-29
Expose spec in cyclonedx-bom


RossComputerGuy:
I'm needing to serialize & deserialize things because the `cached` crate uses serde. I am using `cached` so I can increase performance of some heavy SBOM generation by leveraging caching. There's 2 options which I can see be done. One is make everything in `cyclonedx-bom/src/specs` become public outside of the crate. The other is to make everything have the `serde::Serialize` and `serde::Deserialize` traits. Either option would work but I'd prefer the latter so I don't have to convert between things. If there's another option, I'm all ears but I'm wondering what's the best solution here.

PR #868, opened 2026-05-15
Add support for serde traits to `cyclonedx-bom` types

Naxdy:
Though the intended way to interact with the crate is to use the individual serialization functions for the different specs, sometimes it is desirable to parse into the main types directly when working with custom intermediate formats, or to only (de-)serialize part of a BOM (e.g. only `vulnerabilities`).

This PR adds optional `Serialize` and `Deserialize` derives to the main `Bom` struct and all others that require it, but only when compiling with the `serde` feature. The `serde` feature is opt-in, so as to preserve the current default behavior, and avoid potentially confusing new users with the intended usage.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1162,6 +1162,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f4779c6901a562440c3786d08192c6fbda7c1c2060edd10006b05ee35d10f2d"
 dependencies = [
  "num-traits",
+ "rand",
+ "serde",
 ]
 
 [[package]]
@@ -1374,6 +1376,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34af8d1a0e25924bc5b7c43c079c942339d8f0a8b57c39049bef581b46327404"
 dependencies = [
  "rand_core",
+ "serde",
 ]
 
 [[package]]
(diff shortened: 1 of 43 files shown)


Shnatsel:
Could you explain the motivation behind this change?

This looks like a different take on https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/802 but I'd like to hear what your use cases are like.

Naxdy:
Our use case is an SBOM generation tool that supports merging the final BOM with manual annotations. Currently this includes VEX metadata and subcomponents. Our data sources output partial data that we would like to deserialize into the `cyclonedx_bom` types and merge them with the types we construct within our Rust code.

Currently, we're using our own newtypes and have `impl From<OurType> for CycloneDxType` blocks for every corresponding type. As you can imagine, this is quite the effort to maintain.

The alternative would be to parse into the parent `Bom` type directly, but that would require us to extract the subtypes we care about and perform a lot of unnecessary pattern matching around `Option`s when we know we won't need it.

Basically, we'd like to be able to do something like this:

```rust
#[derive(Serialize, Deserialize)]
struct CustomMeta {
  vulnerabilities: Vec<Vulnerability>,
  subcomponents: Vec<Component>,
  // ...
}
```

---

As a side note, @RossComputerGuy is a colleague of mine, and the SBOM tool referenced in #802 and in this comment are one and the same.

PR #869, opened 2026-05-18
Use `response` instead of `responses` when serializing `VulnerabilityAnalysis`

Naxdy:
The spec actually names this `response` (singular) as opposed to `responses` (plural). Tools like Dependency Track also use the singular `response`, despite the fact that it describes an array.

See: https://cyclonedx.org/docs/1.5/json/#vulnerabilities_items_analysis_response


Shnatsel:
Ah, that's a typo on our end. Thanks!

Naxdy:
I only tested with JSON (not XML), but the tests pass at least

PR #870, opened 2026-06-10
Add --workspace-sboms flag for workspace SBOMs

salim-runsafe:
When `cargo cyclonedx` runs against a rust workspace, it currently emits an SBOM for every workspace member. Some workflows only need the SBOM for the crate being built from a specific manifest.

Add --workspace-sboms with values `all` (default) and `manifest-only`. The default preserves existing behavior. `manifest-only` limits output to the manifest from `--manifest-path` or the current directory.

Return a clear error when `manifest-only` is used with a virtual workspace root manifest that does not match any member crate.

We've been using a fork of this repo with these changes (without the CLI flag) for a while now without issue and wanted to bring the feature upstream.

I wasn't sure if it is appropriate to fill out the `CHANGELOG.md` with an `Unreleased` section or not and can revert that if needed.

### DIFF
--- a/cargo-cyclonedx/CHANGELOG.md
+++ b/cargo-cyclonedx/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+ - Added `--workspace-sboms` flag to control whether SBOMs are generated for every workspace member (`all`, default) or only for the manifest specified by `--manifest-path` or the current directory (`manifest-only`).
+
 ## 0.5.9 - 2026-03-19
 
 ### Added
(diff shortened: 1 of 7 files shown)


Shnatsel:
Just to make sure I understand this correctly: you want to emit a SBOM for only a single specific crate from the workspace, right?

Why are the other crates in the workspace an issue?

salim-runsafe:
Our use case is that we identify `Cargo.toml`s first and then iterate over them while specifying `--override-filename`. This allows us to identify the generated SBOMs amongst other files in a repo with certainty.

With the current upstream behavior this iteration generates the same SBOMs multiple times, once for each crate in the workspace.

Another solution to this problem would be if we could format the output of the command to something structured. Then we could parse the list of generated SBOMs and retrieve them.

Shnatsel:
Ah, it's the "match SBOM to the crate" problem.

Can we write the manifest path (path to Cargo.toml) into the SBOM file? Does that help?

salim-runsafe:
We would still need to be able to know which SBOMs were generated in a structured way, otherwise we're hunting JSON files for that manifest path to try and find a match.

We also need to account for monorepos with multiple unconnected cargo workspaces within them. Today in our fork we can search an entire directory for all Cargo.tomls and then generate SBOMs 1:1 with those tomls and know what we generated because we're specifying the filename with `--override-filename`. The workspaces end up becoming irrelevant to our purposes, which is nice because we don't need to have any special logic for them. We also don't have to worry about side effects generating files we then need to try and sift (syft?!) through to locate the correct SBOMs.

If we were to rely on the current upstream behavior we would end up having to come up with some clever parsing to distinguish workspaces from one another, then generate SBOMs per workspace, then identify the generated SBOMs and map them to each Cargo.toml which they describe.

salim-runsafe:
@Shnatsel given our constraints, what do you think about the proposed solution?

PR #874, opened 2026-07-24
fix #873: Handle binary licenses

### LINKED ISSUE #873: Allow parsing binary files pointed by Cargo's `license-file`
Started [a conversation over Slack](https://cyclonedx.slack.com/archives/C011P6LT8KF/p1784885400059209) to determine if this is a good idea, but perhaps Github Issues is more actively and better for technical discussion.

## Context

I'm running into an issue where my `license-file` is a PDF file for a Proprietary license. This is to be distributed to external users, who might want to use `cargo-cyclonedx` as part of their own SBOM generation.

Because `cargo-cyclonedx` currently doesn't handle non-text license files, I get some warnings related to [this code](https://github.com/CycloneDX/cyclonedx-rust-cargo/blob/d5c35fdf2150c73e2e3570c59fb40daa1f0a96c3/cargo-cyclonedx/src/generator.rs#L462-L485), and the license is never added.

```
[2026-07-24T11:30:42Z WARN  cargo_cyclonedx::generator] Failed to read license file '{package}' for package ../../LICENSE.pdf: stream did not contain valid UTF-8
```

## Proposals

Because `AttachedText`'s content is Base64-encoded, we can easily store files in-place with the appropriate Content Type. This change is sound because Tools that consume the generated SBOM are themselves in charge of processing the content based on their Content Type.

alejnp:
See #873 for more context.

The approach taken changes [`std::fs::read_to_string`](https://doc.rust-lang.org/std/io/fn.read_to_string.html) for [`std::fs::read`](https://doc.rust-lang.org/std/io/fn.read.html).

This avoids UTF-8 validation on read, effectively by-passing a "is text" check, and instead
attempts to determine the correct Content Type from inspection of its bytes, e.g.

* `bytes.starts_with(b"%PDF")` for PDF file and `application/pdf` Content Type
* `std::str::from_utf8(&bytes).is_ok()` for backwards compatibility of text files and  `plain/text` Content Type

Instead of failing on processing any non-PDF binary file, I've decided to simply mark them as `application/octet-stream`. This approach is future-proof in the sense that any "generic" binary files would be accepted, so long as Cargo doesn't disallow the `license-file` itself.


Shnatsel:
Please don't. See #873 for context.

